Issue #15, opened 2017-12-09
Segfault when producing a lot of messages without waiting


thijsc:
To reproduce produce more than ~ 1000 messages and don't wait for them. I think the Ruby GC cleans up the delivery handles which causes librdkafka to crash when it tries to update them.

frsyuki:
We had the same issue. SEGV happens in producer only under high load.
I suspect that the thread of `rd_kafka_poll` loop is trying to fill DeliveryHandle at `Rdkafka::Bindings::DeliveryCallback` and the memory of the DeliveryHandle is already freed by GC because reference to the DeliveryHandle is lost. Reference is lost when `Rdkafka::Producer#produce` raises an exception, or `DeliveryHandle#wait` finishes (successfully or with an error) but then Kafka sends another message.

thijsc:
Indeed. I haven't found a great way to prevent this yet. We run this under high load, but we add all delivery handles to an array and wait for them. So we don't see this behaviour.

I think there will need to be an internal array that the handles are added to and cleared when the delivery handle finishes.

thijsc:
I've reproduced it in this test:

```ruby
it "should produce message that aren't waited for and not crash" do
  100_000.times do
    producer.produce(
      topic:   "produce_test_topic",
      payload: "payload not waiting",
      key:     "key not waiting"
    )
  end

  # Allow some time for a GC run
  sleep 2
end
```

crashes like this most of the time:

```
rspec spec/rdkafka/producer_spec.rb:163(96098,0x70000e32a000) malloc: *** error for object 0x7fe5c98c84f0: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
[1]    96098 abort      bundle exec rspec spec/rdkafka/producer_spec.rb:163
```

Adding the handle to an array fixes the crash. I'm looking into a way to add this mechanism to the producer internally.

mensfeld:
Even with the fix from the commit I still experienced this issue when running several producers in a spec and not closing them fast enough after the specs were done. I fixed that by defining finalizer hooks to close the client before GC kicks in:

```ruby
ObjectSpace.define_finalizer(self, proc { close })
```

ref: https://github.com/karafka/waterdrop/pull/135/files

thijsc:
It seems like ffi's hooks do not always trigger exactly as I expect. We should maybe switch to this approach altogether.

mensfeld:
[...]

PR #16, opened 2018-01-12
Keep delivery_handle until callback touches it

nurse:
delivery_handle need to be kept allocated from Producer#produce to
DeliveryCallback. But since FFI::Struct enables autorelease by default,
it may be released before DeliveryCallback. This causes
DeliveryCallback to write already freed (and maybe used by someone
others) area. It means a memory corruption and it will cause SEGV.

This keeps a reference for the handler to prevent GC.

fix https://github.com/appsignal/rdkafka-ruby/issues/15


thijsc:
Really nice solution, thank you!

I'm looking into why the tests are not passing, we can merge this once the build is green.

thijsc:
Merged this into separate branch, I'll work on getting specs done and green for this scenario.

PR #17, opened 2018-01-15
Keep delivery handle from gc

thijsc:


thijsc:
Continuation of #16

thijsc:
@nurse Looks like this approach does not work in the current incarnation. Multiple calls to `to_ptr` create different pointers, so we cannot reliably use their address as the key in the registry. The registry approach is fine in principle, but we need to find a way to keep the key consistent.

thijsc:
So my diagnosis was slightly wrong. `to_ptr` does return a pointer to the same address. It turns out that the addres needed to be used in the lookup. Fixed this bug and added tests for all this. If the build is green now this should be good to merge.

nurse:
Ah, `message[:_private]` is pointer, not address. Thank you for fix.
CI is failed but it seems just because of travis's temporary failure.

thijsc:
There was a misconfiguration in the CodeClimate config and now the builds don't seem to be exiting properly. Trying some fixes.

thijsc:
Merging, will do a gem release later today.

kzk:
Thank you from Treasure Data team! (CC/ @frsyuki and @nurse)

thijsc:
You're welcome!

PR #31, opened 2018-05-02
Add suport for storing message offsets

### LINKED ISSUE #25: Support manual offset store
When using autocommit, but not auto offset store you can determine when a message is done. `rd_kafka_offset_store` in librdkafka.

thijsc:
Fixes #25. Work in progress.

### DIFF
--- a/lib/rdkafka/bindings.rb
+++ b/lib/rdkafka/bindings.rb
@@ -39,6 +39,7 @@ module Rdkafka
     attach_function :rd_kafka_message_destroy, [:pointer], :void
     attach_function :rd_kafka_message_timestamp, [:pointer, :pointer], :int64
     attach_function :rd_kafka_topic_new, [:pointer, :string, :pointer], :pointer
+    attach_function :rd_kafka_topic_destroy, [:pointer], :pointer
     attach_function :rd_kafka_topic_name, [:pointer], :string
 
     # TopicPartition ad TopicPartitionList structs
@@ -145,6 +146,7 @@ module Rdkafka
     attach_function :rd_kafka_poll_set_consumer, [:pointer], :void
     attach_function :rd_kafka_consumer_poll, [:pointer, :int], :pointer, blocking: true
     attach_function :rd_kafka_consumer_close, [:pointer], :void, blocking: true
+    attach_function :rd_kafka_offset_store, [:pointer, :int32, :int64], :int
 
     # Stats
 
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -187,6 +187,37 @@ module Rdkafka
       out
     end
 
+    # Store offset of a message to be used in the next commit of this consumer
+    #
+    # When using this `enable.auto.offset.store` should be set to `false` in the config.
+    #
+    # @param message [Rdkafka::Consumer::Message] The message which offset will be stored
+    #
+    # @raise [RdkafkaError] When storing the offset fails
+    #
+    # @return [nil]
+    def store_offset(message)
+      # rd_kafka_offset_store is one of the few calls that does not support
+      # a string as the topic, so create a native topic for it.
+      native_topic = Rdkafka::Bindings.rd_kafka_topic_new(
+        @native_kafka,
+        message.topic,
+        nil
+      )
+      response = Rdkafka::Bindings.rd_kafka_offset_store(
+        native_topic,
+        message.partition,
+        message.offset
+      )
+      if response != 0
+        raise Rdkafka::RdkafkaError.new(response)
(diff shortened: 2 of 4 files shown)


mmustala:
How far this is from getting merged to master? We really need it. Anything I can do for it?

thijsc:
Close! The code is ok and works locally. I've been having a few issues with the Travis build that's holding up this pull. Will try to fix it this week.

thijsc:
I'm seeing an assert failure error when calling `store_offset`. There's some discussion here: https://github.com/edenhill/librdkafka/issues/1000

Also trying to debug it.

thijsc:
Update on this: I'm still struggling to write truly reliable tests for this. The implementation code itself works fine.

mjkillough:
Thanks a lot for this @thijsc! We have an application where we'd like to use `store_offset`.

I was able to reproduce and debug the intermittent test failures by adding:

```ruby
puts "#{consumer.committed}"
```

... [before the call to `store_offset` in the test](https://github.com/appsignal/rdkafka-ruby/blob/a6f361611e78c3433d666d262c50188777787b8f/spec/rdkafka/consumer_spec.rb#L242).

With this debug logging, I can see that when the test fails for me, the `consumer` is not subscribed to partition 1, which is the partition we're storing the offset for:

```
Run options: include {:locations=>{"./spec/rdkafka/consumer_spec.rb"=>[243]}}
<TopicPartitionList: {"consume_test_topic"=>[<Partition 0 with offset 1>, <Partition 1 with offset 1>]}>
.

Finished in 5.48 seconds (files took 0.37455 seconds to load)
1 example, 0 failures

Coverage report generated for RSpec to /Users/michaelkillough/Code/rdkafka-ruby/coverage. 338 / 444 LOC (76.13%) covered.
Run options: include {:locations=>{"./spec/rdkafka/consumer_spec.rb"=>[243]}}
<TopicPartitionList: {"consume_test_topic"=>[<Partition 0 with offset 1>, <Partition 1 with offset 1>]}>
.

Finished in 5.56 seconds (files took 0.26423 seconds to load)
1 example, 0 failures

Coverage report generated for RSpec to /Users/michaelkillough/Code/rdkafka-ruby/coverage. 338 / 444 LOC (76.13%) covered.
Run options: include {:locations=>{"./spec/rdkafka/consumer_spec.rb"=>[243]}}
<TopicPartitionList: {"consume_test_topic"=>[<Partition 2 with offset 1>]}>
F

Failures:

  1) Rdkafka::Consumer#commit, #committed and #store_offset with a commited consumer #store_offset should store the offset for a message
     Failure/Error: raise Rdkafka::RdkafkaError.new(response)
     
     Rdkafka::RdkafkaError:
       Local: No offset stored (no_offset)
     # ./lib/rdkafka/consumer.rb:239:in `commit'
     # ./spec/rdkafka/consumer_spec.rb:244:in `block (5 levels) in <top (required)>'

Finished in 5.32 seconds (files took 0.36466 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/rdkafka/consumer_spec.rb:236 # Rdkafka::Consumer#commit, #committed and #store_offset with a commited consumer #store_offset should store the offset for a message
```

It seems the `let!(:message)` binding is creating a second consumer within the same consumer group, which means it's sometimes unlucky and partition 1 is assigned to the consumer created for `message`. (As far as I can tell, `store_offsets` will silently allow you to store an offset for a partition you aren't assigned to, but will ignore it when later committing).

I'm not sure why the `message` is defined with `let!`, but the tests seem to pass reliably when I change it to `let`. 

I think you can either:
 - Switch the `let!` to `let`.
 - Alter `wait_for_message` to accept a `consumer`, rather than a `config`, so that the group only ever has one consumer in it.
 - Alter the `store_offset` test to store offsets for a message its consumer actually receives, rather than a `double()`.

What do you think? I'd be happy to try come up with a change to do one of the above.

thijsc:
Thanks for digging in! Was a bit burnt out on this issue. I think this solution is probably smartest:

> Alter wait_for_message to accept a consumer, rather than a config, so that the group only ever has one consumer in it.

mjkillough:
No worries! Intermittent test failures are a pain to debug. :-) Thanks a lot for the quick response.

> I think this solution is probably smartest:
>> Alter wait_for_message to accept a consumer, rather than a config, so that the group only ever has one consumer in it.

I'm happy to give this a go.

mjkillough:
I went with all three options and submitted a PR against your branch: #51.

I've run the spec 50 times locally and don't get any failures. (I was previously getting failures every 4 runs).

thijsc:
Thanks for your help @mjkillough!

Issue #42, opened 2018-09-14
Random installation failures


Numbluk:
I was attempting to install `rdkafka-ruby` with `gem install rdkafka -v '0.3.5'`. It failed several times in a row with:

```
Building native extensions. This could take a while...
ERROR:  Error installing rdkafka:
	ERROR: Failed to build gem native extension.

    current directory: /home/lukas/.rvm/gems/ruby-2.5.0/gems/rdkafka-0.3.5/ext
/usr/share/rvm/rubies/ruby-2.5.0/bin/ruby -rrubygems /home/lukas/.rvm/gems/ruby-2.5.0@global/gems/rake-12.3.0/exe/rake RUBYARCHDIR=/home/lukas/.rvm/gems/ruby-2.5.0/extensions/x86_64-linux/2.5.0/rdkafka-0.3.5 RUBYLIBDIR=/home/lukas/.rvm/gems/ruby-2.5.0/extensions/x86_64-linux/2.5.0/rdkafka-0.3.5
2 retrie(s) left for v0.11.3.tar.gz
1 retrie(s) left for v0.11.3.tar.gz
0 retrie(s) left for v0.11.3.tar.gz
nil can't be coerced into Integer
Extracting v0.11.3.tar.gz into tmp/x86_64-linux-gnu/ports/librdkafka/0.11.3... ERROR, review '/home/lukas/.rvm/gems/ruby-2.5.0/gems/rdkafka-0.3.5/ext/tmp/x86_64-linux-gnu/ports/librdkafka/0.11.3/extract.log' to see what happened. Last lines are:
========================================================================
tar (child): ports/archives/v0.11.3.tar.gz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
========================================================================
rake aborted!
Failed to complete extract task
/home/lukas/.rvm/gems/ruby-2.5.0/gems/mini_portile2-2.2.0/lib/mini_portile2/mini_portile.rb:400:in `block in execute'
/home/lukas/.rvm/gems/ruby-2.5.0/gems/mini_portile2-2.2.0/lib/mini_portile2/mini_portile.rb:371:in `chdir'
/home/lukas/.rvm/gems/ruby-2.5.0/gems/mini_portile2-2.2.0/lib/mini_portile2/mini_portile.rb:371:in `execute'
[...]

PR #43, opened 2018-09-15
Update remote location of 'librdkafka' tarball to avoid GitHub rate limits

koenrh:
In a shared (cloud) environment (e.g. [Google Cloud Build](https://cloud.google.com/cloud-build/), a container builder) you hit GitHub's rate limit fairly quickly. To work around these rate limits you could fetch resources directly from GitHub's 'codeload', which has a less stringent rate limit. Requests to `github.com` (as opposed to `codeload.github.com`) seem to be routed through some authentication proxy, which might explain the more stringent rate limits.

### DIFF
--- a/ext/Rakefile
+++ b/ext/Rakefile
@@ -3,6 +3,22 @@ require "mini_portile2"
 require "fileutils"
 
 task :default => :clean do
+  # MiniPortile#download_file_http is a monkey patch that removes the download
+  # progress indicator. This indicator relies on the 'Content Length' response
+  # headers, which is not set by GitHub
+  class MiniPortile
+    def download_file_http(url, full_path, _count)
+      filename = File.basename(full_path)
+      with_tempfile(filename, full_path) do |temp_file|
+        params = { 'Accept-Encoding' => 'identity' }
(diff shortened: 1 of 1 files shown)


thijsc:
It seems to be a good idea to do this because the original URL redirects here anyway. Travis is still having trouble reliably downloading though. Retrying the build.

koenrh:
This is a different issue. GitHub no longer appears to honor the 'Accept-Encoding: identity' ("please do not encode responses") header, which means that no 'Content-Length' is set on responses. 'Mini portile' relies on this header to display a download progress indicator.

This might be the cause of the issue others have reported in https://github.com/appsignal/rdkafka-ruby/issues/42.

I'm not too fond of monkey patching things, but I have added one anyway in 9d04118 as a quick fix to work around this particular issue. I'll file issues with 'mini portal' and GitHub for the aforementioned issues.

koenrh:
This issue was already reported: https://github.com/flavorjones/mini_portile/issues/85

niels-s:
Any news on this PR? Would like to get build images on Google working again 🙏

thijsc:
This is good to go now, merging. I'll get a release out today or tomorrow.

PR #46, opened 2018-10-02
Add gem tasks to main Rakefile

MadBomber:
I just added the bundler/gem_tasks to the main Rakefile so that I could build and install the gem locally.


thijsc:
Thanks!

PR #49, opened 2018-10-19
Add missing runtime dependency to rake

aboutNisblee:


thijsc:
Hm is it because we use Rake for the install? I might have missed that since I only use this gem in projects that already have Rake as a dependency.

aboutNisblee:
The librdkafka build process is started by a Rakefile. If your project doesn't already include rake in its dependencies, the build will fail until you explicitly install rake.

thijsc:
Makes sense

PR #51, opened 2018-10-31
store_offset: pass actual message in tests.

mjkillough:
Pass an actual message to `Consumer#store_offset` in the test. Otherwise
we might try to commit an offset for partition that our consumer isn't
subscribed to.

This changes `wait_for_message` to accept a `consumer`, so that we can
receive and `#store_offset` from the same consumer.

### DIFF
--- a/spec/rdkafka/consumer_spec.rb
+++ b/spec/rdkafka/consumer_spec.rb
@@ -123,13 +123,11 @@ describe Rdkafka::Consumer do
       ).wait
     end
 
-    # Wait for message commits the current state,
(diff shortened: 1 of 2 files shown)


spec/rdkafka/consumer_spec.rb
```diff
@@ -123,13 +123,11 @@
       ).wait
     end
 
-    # Wait for message commits the current state,
-    # commit is therefore tested here.
```
mjkillough: I removed this comment as it was no longer true.

PR #53, opened 2018-11-12
Fix documentation link

GregMefford:
I noticed that the existing link was pointing to essentially the same thing as the `README`, but with broken CSS.


thijsc:
Thanks!

PR #62, opened 2018-11-22
Document list param of commit method

niels-s:
When I was browsing through the source code I noticed the documentation was missing for the list param of the commit method of a consumer.

### DIFF
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -223,6 +223,7 @@ module Rdkafka
 
     # Commit the current offsets of this consumer
     #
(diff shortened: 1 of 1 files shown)


lib/rdkafka/consumer.rb
```diff
@@ -223,6 +223,7 @@ def store_offset(message)
 
     # Commit the current offsets of this consumer
     #
+    # @param list [TopicPartitionList] The topic with partitions to commit
```
thijsc:
```suggestion
    # @param list [TopicPartitionList,nil] The topic with partitions to commit
```

Issue #57, opened 2018-11-14
Poll in batches


JHK:
Handling the offset right is cumbersome. Especially, when I want to consume a few messages, do a batch-insert into the database and don't want an offset change before the batch-insert has succeeded. To do so I've disabled the `auto.commit` and do the `commit` manually after the batch got processed successfully (see https://github.com/zendesk/racecar/commit/d00dbc733680318b65ba2b5e28e36a7cfec81086).

This can probably be implemented using the [`rd_kafka_consume_batch`](https://docs.confluent.io/3.2.1/clients/librdkafka/rdkafka_8h.html#a53511739a2cf498b8d88287fef6873ce) binding, which allows to auto commit *after* a callback got processed successfully.

mensfeld:
I'm looking into switching to rdkafka as well once some of the things are available. What I also lack is the batch metadata details (see here: https://github.com/karafka/karafka/blob/1.3-wip/lib/karafka/connection/batch_delegator.rb#L26 and here: https://github.com/karafka/karafka/blob/1.3-wip/lib/karafka/params/builders/metadata.rb)

mensfeld:
Just as a clarification on the purpose of that: it is just easier to implement rejection logic based on the offset lag for example (so old things won't be processed at all when the lag is growing in favour of the new once)

thijsc:
@Paxa asked a question related to this issue:

> What non-legacy API can we use to implement poll in batches?

I just looked into the docs and there is no poll in batches in the new consumer API. I think this is the case because there is no technical need for it. The high-level consumer already batches under the hood. You're getting messages delivered one by one, but they're being popped of internal batches.

I do think there is a use case we have to think about though, people seem to use the batches to make handling commits easier. @JHK said:

> Handling the offset right is cumbersome. Especially, when I want to consume a few messages, do a batch-insert into the database and don't want an offset change before the batch-insert has succeeded. To do so I've disabled the auto.commit and do the commit manually after the batch got processed successfully (see zendesk/racecar@d00dbc7).

I would implement this using [`store_offset`](https://www.rubydoc.info/github/appsignal/rdkafka-ruby/Rdkafka/Consumer#store_offset-instance_method). You can disable `enable.auto.offset.store` and use this method to control the exact offset that will be stored on the next commit.

@JHK Does that fix this use case for you?

thijsc:
@JHK Do you think using store offset would support your use case?

JHK:
[...]

PR #68, opened 2019-02-27
Add Consumer#consume_with_queue

Paxa:
* Add Consumer#consume_with_queue
* Add Consumer#fetch_partitions_count
* Add Consumer#reset_offsets
* Add benchmarks
* Change require to require_relative
* Add Rdkafka.print_ffi_obj for debugging

`consume_with_queue` is about 20-30% faster in benchmarks and hopefully be more efficient with real loads. Will test with real app later and show numbers

---

Example of output by `Rdkafka.print_ffi_obj`
```
Rdkafka::Bindings::KafkaMetadata
* broker_cnt: 1
* brokers: #<FFI::Pointer address=0x00007f7fa958c478>
* topic_cnt: 1
* topics: #<FFI::Pointer address=0x00007f7fa958c4a0>
* orig_broker_id: 1001
* orig_broker_name: "localhost:9092/1001"

Rdkafka::Bindings::KafkaMetadataTopic
* topic: "rake_test_topic"
* partition_cnt: 3
* partitions: #<FFI::Pointer address=0x00007f7fa958c4d0>
* err: 0
```

---

I could not implement reading in batches, it gives me array of messages but I'm not sure how to handle it correctly with ffi


Paxa:
With real apps I see improvements by 15-25%, less significant than I expected

thijsc:
Hi Paxa, thanks for your work on this!

I'm always open to good arguments that might convince me otherwise, but supporting the legacy consumer is not something I want to add. I talked about this a bit in the readme:

> This gem only provides a high-level Kafka consumer. If you are running an older version of Kafka and/or need the legacy simple consumer we suggest using the Hermann gem.

Our purpose with this gem was providing a production-ready way for people to use all the modern Kafka features. It is explicitly not a goal to support the old way of doing things. If you want the old way I think Hermann is a perfectly fine alternative to this gem.

So anything that's marked as legacy in rdkafka's doc is not going to be supported here. I think that's the best course to keep this project focussed and maintainable. So I'm afraid I'm not going to merge this change even if it were completely finished. Do you have any great arguments that you think need to be made?

Paxa:
Just though it will have better performance if process in batches, but I could not use batch api.
I started to look into it because my consumer works slower then I expected, I see pauses between processing several messages.

Other things in this PR are just useful to run examples and benchmarks.

About changing `require` to `require_relative` - rake tasks doesn't work with usual require

Paxa:
Btw how do we know which functions are deprecated?

thijsc:
We do have an issue open to poll in batches btw: #57.

I mean "legacy" instead of deprecated. Most of rdkafka is not, but the simple consumer is. Did you try just assigning and not subscribing with the high level consumer btw? Maybe that's as fast as the simple consumer since it does the same approach.

Paxa:
What non-legacy API can we use to implement poll in batches?

thijsc:
> What non-legacy API can we use to implement poll in batches?

Good question, will answer over in #57

PR #69, opened 2019-03-12
Implement pause/resume partitions

### LINKED ISSUE #55: Ability to pause partitions
When there is an business logic error (e.g. due to unexpected message content) I would like to make progress at least on other partitions. Therefore I propose the following pseudo-code as a draft, implementing the `pause` method on `Rdkafka::Consumer`:

```ruby
loop do
  message = consumer.poll
  process(message) if message
rescue Rdkafka::RdkafkaError => e
  raise unless e.is_partition_eof?
rescue => e
  consumer.pause(message.topic, message.partition)
end
```

For reference:
 * [Kafka consumer docs on `pause`](https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#pause-java.util.Collection-)
 * [corresponding `librdkafka` binding](https://docs.confluent.io/3.2.1/clients/librdkafka/rdkafka_8h.html#a293cb2aac290c157702d3b82f5c14fce)

dmexe:
Add bindings to:

- rd_kafka_pause_partitions
- rd_kafka_resume_partitions

Fixes #55

### DIFF
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -64,7 +64,7 @@ module Rdkafka
     #
     # @param list [TopicPartitionList] The topic with partitions to pause
     #
-    # @raise [RdkafkaError] When pausing subscription fails.
+    # @raise [RdkafkaTopicPartitionListError] When pausing subscription fails.
     #
     # @return [nil]
     def pause(list)
@@ -74,7 +74,8 @@ module Rdkafka
       tpl = list.to_native_tpl
       response = Rdkafka::Bindings.rd_kafka_pause_partitions(@native_kafka, tpl)
       if response != 0
-        raise Rdkafka::RdkafkaError.new(response, "Error pausing '#{list.to_h}'")
+        list = TopicPartitionList.from_native_tpl(tpl)
+        raise Rdkafka::RdkafkaTopicPartitionListError.new(response, list, "Error pausing '#{list.to_h}'")
       end
     end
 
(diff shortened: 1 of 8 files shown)


lib/rdkafka/consumer.rb
```diff
@@ -60,6 +60,42 @@ def unsubscribe
...
+    #
+    # @return [nil]
+    def resume(list)
+      unless list.is_a?(TopicPartitionList)
+        raise TypeError.new("list has to be a TopicPartitionList")
+      end
+      tpl = list.to_native_tpl
+      response = Rdkafka::Bindings.rd_kafka_resume_partitions(@native_kafka, tpl)
+      if response != 0
+        raise Rdkafka::RdkafkaError.new(response, "Error resume '#{list.to_h}'")
```
thijsc: I'd like to add an optional list field to the error class, so you can put the actual object in there for further inspection.

thijsc: Or maybe subclass error to `RdkafkaErrorWithTpl` or something like that.

dmexe: I've carefully looked at `TopicPartitionList` and I think  the first we need is a flattened  representation of `TopicPartitionList` which would be the array of TopicPartition structures

```
class TopicPartition
  attr :topic
  attr :partition
  optional :offset
  optional :err_t
end
```

So, per a partition error could be represented as `Hash<TopicPartition, RdkafkaError>`, which could be used inside `RdkafkaErrorWithTpl `

thijsc: That would indeed be nice, I'm not a huge fan of how rdkafka combines so many pieces of information in a single struct. Would make sense to create a dedicated object.

PR #71, opened 2019-03-18
Support message headers for consumer & producer

dmexe:
* add field `Consumer::Message#headers`
* add `headers` argument to `Producer#delivery`


mensfeld:
Hey, it seems that the docs: https://www.rubydoc.info/github/appsignal/rdkafka-ruby/Rdkafka/Producer don't reflect this feature.

PR #74, opened 2019-03-25
Fix memory leak for TopicPartitionList#to_native_tpl while calling consumer's pause/resume.

dmexe:
* `TopicPartitionList#to_native_tpl` returns a `FFI::AutoPointer` which will be cleaned by `rd_kafka_topic_partition_list_destroy`
* doesn't release pointer in `TopicPartitoinList#from_native_tpl`, releasing must be at the same level as pointer allocation.

### DIFF
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -30,7 +30,8 @@ module Rdkafka
     # @return [nil]
     def subscribe(*topics)
       # Create topic partition list with topics and no partition set
-      tpl = Rdkafka::Bindings.rd_kafka_topic_partition_list_new(topics.length)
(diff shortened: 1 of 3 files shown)


lib/rdkafka/consumer.rb
```diff
@@ -73,6 +71,7 @@ def pause(list)
       end
       tpl = list.to_native_tpl
       response = Rdkafka::Bindings.rd_kafka_pause_partitions(@native_kafka, tpl)
+
       if response != 0
         list = TopicPartitionList.from_native_tpl(tpl)
```
dmexe: here was a memory leak, it doesn't properly clean a pointer if response code is 0

PR #75, opened 2019-03-28
bump librdkafka to 1.0.0

breunigs:
I didn't bump rdkafka's version number with this patch, but since librdkafka has breaking changes, I assume a major version would be well suited:
https://github.com/edenhill/librdkafka/releases/tag/v1.0.0

We've been running librdkafka 1.0-RC7 with rdkafka-ruby 24e5a537a8c5fae02981a6f49cb8cae0cd9c5e2e so far, which works nicely enough. I'll retest this with librdkafka's final release and rdkafka-ruby's tip and report if I see any issues.


thijsc:
Thanks! Where did you get the checksum from? Travis is buggy today, will try to restart the build.

breunigs:
The checksum is from downloading the release tar ball from github with different ISPs. I don't think upstream provides the verified checksums anywhere, although all the tooling and release instructions were merged some time ago: https://github.com/edenhill/librdkafka/commit/5140cd9d965406a144de067b775bc9ca5255aea1

breunigs:
Hm, the timeout being shown on Travis is not random, I experience it locally as well. It consistently fails with `1000ms` and below. It works sometimes with `1001ms` and always on my machine with `1100ms`. It seems `rd_kafka_committed` gained a 1s sleep somewhere…

breunigs:
I bisected it to https://github.com/edenhill/librdkafka/commit/382bfd8afe9702cceafe88913d5d77340fff3c60 and it's fast after the initial call, so I guess it requires time to setup a new connection. Increasing the timeout or leading with a call with a high enough timeout fixes the test failure, but I wonder if the default timeout should be bumped as well.

thijsc:
> Increasing the timeout or leading with a call with a high enough timeout fixes the test failure, but I wonder if the default timeout should be bumped as well.

I've often noticed that these timeouts occur much less often in real-world scenarios because existing connections are used there. I think I would indeed just increase the default timeout here a bit.

mjkillough:
This looks great! It would be awesome to see this make its way into a new release of the gem. Is there anything remaining that I could help out with to get it there?

We'd like to upgrade some of our services to run `librdkafka` 1.0.0 as the new connection management seems to resolve a few problems we've been having with disconnects.

thijsc:
Build is green now, so good to go!

mjkillough:
> Build is green now, so good to go!

Amazing!! Thank you so much for the quick turnaround. ❤️

thijsc:
I just pushed 0.5.0 with the latest changes.

PR #81, opened 2019-08-08
Some styling and quality remarks

mensfeld:
During my work on porting waterdrop to rdkafka, I looked into the sources and I liked what I saw ;)

here are some remarks from coditsu that fix some small typos and add some remarks.

Things resolved: https://app.coditsu.io/karafka/builds/diffs/diffs/e9fdb822-e52c-453e-8b76-3d799c431575/resolved


mensfeld:
Also I do believe we should allow non int values for the max timeout for a more granular control (happy to add it).

thijsc:
Thanks!

PR #84, opened 2019-08-17
Get consumer position

### LINKED ISSUE #24: Get position of a consumer
Support getting the position of a consumer, `rd_kafka_position` in librdkafka. Result should be supported in `lag`.

thijsc:
Fixes #24


mensfeld:
@thijsc do you recall why this never reached master?

thijsc:
I think this got stuck because the build was failing for months. Should work as-is.

PR #94, opened 2019-11-02
fix travis builds

nijikon:


mensfeld:
Irony: now all fail :D which is good - at least a bit of predictability :dancer:

rgo:
Locally, if I run consumer_spec.rb it consistently fails (Assertion failed error) in #seek subscription:

```
Rdkafka::Consumer
  #subscripe, #unsubscribe and #subscription
    should subscribe, unsubscribe and return the subscription
    should raise an error when subscribing fails
    should raise an error when unsubscribing fails
    should raise an error when fetching the subscription fails
  #pause and #resume
    should raise when not TopicPartitionList
    should raise an error when pausing fails
    should raise an error when resume fails
    subscription
      should pause and then resume
  #seek
    should raise an error when seeking fails
    subscription
Assertion failed: (r == 0), function rwlock_wrlock, file tinycthread_extra.c, line 138.

[Process exited 134]
```

Running #seek context isolated works.

rgo:
> Irony: now all fail :D which is good - at least a bit of predictability 💃

It seems to be failing because `dockerize` is not installed or something like that:

```
$ dockerize -wait tcp://localhost:2181 -timeout 1m
dockerize: command not found
```
https://travis-ci.org/appsignal/rdkafka-ruby/jobs/606363122#L421

nijikon:
@mensfeld @rgo this is a draft, be patient 😸

nijikon:
@thijsc could you give me some more privileges so I can cancel builds on travis? Right now I can't.

thijsc:
I'm also working on this in #98

thijsc:
Did you get a green build locally by the way?

thijsc:
Thanks for your work on this! The build was fixed in #108 in the end.

Issue #99, opened 2019-11-13
Consume In Batches API


KTSCode:
I know that it's currently possible to consume in batches [poll in batches issue](https://github.com/appsignal/rdkafka-ruby/issues/57); however, I believe the API for this could be significantly improved. One of the main draws of this library was the simplicity with which a Kafka consumer could be implemented. Consuming messages individually has become too slow for my current use case. I began the transition to consuming in batches, and have found the work required to be more challenging than I feel is strictly necessary. 
 
I think the users of this library would find an API similar to the code below far superior to the current interface. 
```ruby
consumer.each_in_batches(<batch size>, <timeout>) do |batch|
  batch.each |message|
     puts "Message received: #{message}"
  end   
end
```
* `batch size` -> The number of messages to be consumed before the block of code is run.
* `timeout` -> The time in milliseconds waited before a batch of greater than 0 messages and less than `batch size` messages is passed to the code block.
* It would be expected that any errors that occur would propagate up and the offset for those messages would not be committed.

thijsc:
Thanks for your feedback on this!

I think it's important to understand that the high-level consumer API that replaces the now legacy simple consumer API does not do batching in the same way. The high-level consumer always works in batches, it just does not expose the workings of that in its public API. Adding some kind of batching layer cannot improve performance, it will even be a little slower because extra work needs to be done to re-batch messages on the Ruby side.

When you consume a message in the naive way you are actually using batches. It is possible to tune the batching behaviour using config options such as `fetch.max.bytes` which are documented [here](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md).

Personally I feel that using `store_offset` and controlling the timing of a commit is the best way to handle this. It maps directly to what the driver is actually doing. If we were to implement a wrapper similar to the example above we'd use those exact techniques underwater since this is the only possible way of achieving this behaviour with the high-level consumer.

In my mind, the argument for doing this would have to be that it's a substantially better API for a lot of people. I'd love to hear more if you think that's the case.

KTSCode:
Thank you for the quick response @thijsc!

I only have my personal experience to go off of, but I would like to make rdkafka-ruby the default ruby Kafka library at the company I work for. I would feel much better telling other teams to use this tool if I knew there was a simple batch consumption API that could be quickly and easily implemented. 

[...]

PR #100, opened 2019-11-13
A Naive Consume in Batches API

KTSCode:
This is an example of how one could naively implement a consume in batches API, however it is lacking support for not committing on error. I assume this could be done much more efficiently though It would take me a lot more time reading the code to figure out how to do that.


thijsc:
Closing as discussed in #99

PR #103, opened 2020-01-03
ext: Add build:git[ref] task for building librdkafka from a git sha

gaffneyc:
As part of some work to look into the build failures (see #98) I wanted
an easy way to run an automated bisect against different librdkafka
commits. This task is designed to be used during development and the
default task should still be used when installing the gem.

Examples:
```console
rake build:git[ff9478c62126]
rake build:git[v1.3.0-RC2]
```


thijsc:
Thanks!

PR #104, opened 2020-01-03
Update Kafka to 2.2.1 for development and testing

gaffneyc:
Main goal was two fold: to update to a newer Kafka version to see if
that would fix some of the test failures and to simplify the config.
I've switched to using `127.0.0.1` instead of `localhost` as I was
getting connection failures using the hostname that may be related to
IPv6, switching to the IP made them disappear. I've also removed
KAFKA_ADVERTISED_PORT as it appears to be the default and
AUTO_CRETE_TOPICS_ENABLE as it looks to have been removed. Finally,
Zookeeper is no longer exposed as Kafka can connect to it within the
Docker network just fine. Linking to the docker socket file doesn't
appear to be necessary in this single broker case.

In testing, I was seeing tests pass just fine with 2.2.1 but failing
with 2.3.1.

Based on part of the work done in #98

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,18 +1,18 @@
-version: '2'
+version: "3"
+
 services:
   zookeeper:
     image: wurstmeister/zookeeper
-    ports:
-      - "2181:2181"
+
   kafka:
-    image: wurstmeister/kafka:1.0.1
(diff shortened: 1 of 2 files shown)


thijsc:
We now use the latest kafka version, was merged in another pull. So closing this one.

PR #107, opened 2020-01-24
Ensure the consumer always destroy native TopicPartitionList

gaffneyc:
When converting a Rdkafka::TopicPartitionList to a native type, Rdkafka
was depending on AutoPointer to eventually free them. There are a couple
cases where a Client handle returns a pointer to the application that
the application is then required to free, which could be leading to test
instability.

This takes a pass at ensuring all uses of native TopicPartitionList
instances (even those return) are deterministically freed.

### DIFF
--- a/spec/rdkafka/consumer_spec.rb
+++ b/spec/rdkafka/consumer_spec.rb
@@ -6,6 +6,9 @@ describe Rdkafka::Consumer do
   let(:consumer) { config.consumer }
   let(:producer) { config.producer }
 
+  after { consumer.close }
+  after { producer.close }
+
   describe "#subscripe, #unsubscribe and #subscription" do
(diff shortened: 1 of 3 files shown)


spec/rdkafka/consumer_spec.rb
```diff
@@ -205,8 +208,6 @@ def send_one_message(val)
         expect(records&.payload).to eq "payload c"
         records = consumer.poll(timeout)
         expect(records).to be_nil
-
-        consumer.commit
```
gaffneyc: Removing the commit got the test passing. It was failing with `Local: No offset stored (no_offset)` which I'm guessing the seek+poll didn't cause the internal offset to be advance or it did but had already committed.

gaffneyc:
On Travis one of the tests passed so this seems like it is a move in the right direction.

PR #114, opened 2020-04-15
Switch to Semaphore CI

thijsc:


mensfeld:
@thijsc can you provide a justification for this decision? I'm curious of the reason to switch from `travis` to `semaphore`. Thanks!

thijsc:
We've been using Semaphore at our business for a while now. We recently switched over our open-source builds too and they've become a factor faster and more consistent.

The test setup here is pretty heavy, so think it might be beneficial for this project too. It's not a given that we will actually switch, seeing if I can get a green build so we can compare. Input is welcome.

thijsc:
Closing for the moment, haven't had time to finish this.

PR #115, opened 2020-04-24
Make sure producer is closed before it is GCed by Ruby

mensfeld:
This PR aims to optimize the problem presented here: https://github.com/appsignal/rdkafka-ruby/issues/15

It makes sure, that before a producer is GCed, it gets closed. I've tested it here: https://github.com/karafka/waterdrop/pull/135

Actions would crash every 3rd build randomly. After we close upon finalization, the problem is no longer present. This is also not a problem for already closed producer as then, nothing happens.

This will ensure, that all the FFI callbacks are handled properly without a segfault.

### DIFF
--- a/lib/rdkafka/producer.rb
+++ b/lib/rdkafka/producer.rb
@@ -11,6 +11,10 @@ module Rdkafka
     def initialize(native_kafka)
       @closing = false
       @native_kafka = native_kafka
+
+      # Makes sure, that the producer gets closed before it gets GCed by Ruby
+      ObjectSpace.define_finalizer(self, proc { close })
+
       # Start thread to poll client for delivery callbacks
       @polling_thread = Thread.new do
         loop do


Adithya-copart:
Looks good to me.

Adithya-copart:
@mensfeld I found something interesting
https://www.mikeperham.com/2010/02/24/the-trouble-with-ruby-finalizers/

You might have to undo the change in https://github.com/karafka/waterdrop/pull/135 as well.
```
class A
  def initialize(val)
    @val = val
    ObjectSpace.define_finalizer(self, proc {reset})
  end

  def reset
    p 'reset'
  end
end

initial = ObjectSpace.each_object(A).count
# => 0

a = A.new(5)
after_initialize = ObjectSpace.each_object(A).count
# => 1

a=nil; GC.start

after_gc = ObjectSpace.each_object(A).count
# => 1
```
We can use a class method as suggested and pass the handle and polling thread as arguments to get around this possible memory leak.

Not related to this PR but one of the other problems I saw while using `AutoPointer` that uses a finalizer is that the underlying socket can be closed in a child process if `Process#fork` is used. We don't have a way to check if the `handle` is still open and is something that could be implemented.

mensfeld:
@Adithya-copart really good catch however I actually don't think I'm going to fix that but rather document. The reason here is, that if you:
a) run it in a short living scripts, doesn't matter
b) run it the way you should, that is with a producer per cluster (or the configuration you need) and you don't open and close the producers all the time (why would you?), this actually is not that harmful.

Majority of the cases that I saw within the Karafka ecosystem are cases when you run things with a single producer in a "singleton like" manner. That's why I made sure it's thread safe. Having an additional thread with Ruby scheduler is not something I want to add unless extremely necessary. However I do appreciate the finding a lot!

I will take the weekend to think about it. Maybe there's a different solution that will make code elegant yet that will satisfy the GC.

mensfeld:
@Adithya-copart after giving it a second shot in WaterDrop I think, I will leave it as it is there for the time being. That is until we fix it here (which I can work in around 7-10 days) - to tackle it as you suggested. ATM unless several producers are constantly created and closed I don't see the risk. It's a pre "not yet released" version and I don't see this problem in the systems where we use it. However if it makes to the final release I'm 100% sure I will document this behavior. 

Thank you once again for pointing that out.

mensfeld:
@Adithya-copart @thijsc as a followup, those are my conclusions:

- The main problem here (at least from my perspective) is segfaults when exiting Ruby (in my case rspec). The finalizer does prevent this as only then it is called (as @Adithya-copart GC won't remove it due to the closure reference during normal sweeping). 
- I would not worry that much about use-cases where someone creates and abandons a lot of producers. That does not make any sense and would not be even close to being sub-optimal
- I would suggest adding an `#undefine_finelizer` usage when the producer is correctly closed (see https://github.com/karafka/waterdrop/pull/137/files#diff-64136f6639419a7d972a619ba564bd7bR90)

I'm fine implementing it as long as we agree on going that way or building the finalizer by passing the handle.

it works, here's a poc:

```ruby
class A
  def initialize(val)
    @val = val
    ObjectSpace.define_finalizer(
      self,
      proc { close }
    )
  end

  def close
    p 'closed'
    ObjectSpace.undefine_finalizer self
  end
end

initial = ObjectSpace.each_object(A).count
# => 0
a = A.new(5)

p after_initialize = ObjectSpace.each_object(A).count
# => 1

a.close

a=nil; GC.start

p after_gc = ObjectSpace.each_object(A).count #=> 0
```

> Not related to this PR but one of the other problems I saw while using AutoPointer that uses a finalizer is that the underlying socket can be closed in a child process if Process#fork is used. We don't have a way to check if the handle is still open and is something that could be implemented.

I fixed that as well here: https://github.com/karafka/waterdrop/pull/137

I simply don't allow to use producers that were already in use in the parent process.

Adithya-copart:
> I would suggest adding an #undefine_finalizer usage when the producer is correctly closed

Nice, I think this approach along with a finalizer that does not reference the object directly will solve the problem of segmentation faults or related errors and GC the producers that are not explicitly closed.

I also agree that instantiating and not closing producers may not be part of a real-world application. Since there is a possible solution, let's address this use-case as well.

thijsc:
Sorry for the slow reply!

This looks good to me. Wouldn't adding `return if @closing` to `close` fix a potential double call too?

Adithya-copart:
@thijsc The change is in the `initialize` constructor, so `return if @closing` is really not necessary.

> https://www.mikeperham.com/2010/02/24/the-trouble-with-ruby-finalizers/

The finalizer as-is will never be called as mentioned in the blog post, so we can probably skip adding this altogether.

PR #118, opened 2020-05-04
Semaphore 2

Adithya-copart:
Take on https://github.com/appsignal/rdkafka-ruby/pull/114


Adithya-copart:
@thijsc Not sure how I can trigger the CI pipeline to test this.
The issue in https://github.com/appsignal/rdkafka-ruby/pull/114 is that Semaphore only caches `vendor/bundle` and the compiled C-extension is missing.

thijsc:
Thanks, I've added your change in #114

PR #120, opened 2020-05-26
Fix metadata.rb references to RdkafkaError

Adithya-copart:
Fix references from `Rdkafka::Error` to `RdkafkaError`.

PR #122, opened 2020-06-15
fix uninitialized constant Rdkafka::Error

archfish:


archfish:
quick fix：

```ruby
module Rdkafka
  class Error < RdkafkaError; end
end
```

Adithya-copart:
Duplicate of https://github.com/appsignal/rdkafka-ruby/pull/120

We can define `Rdkafka::Error` like this PR or switch to `Rdkafka::RdkafkaError` to be consistent everywhere.

PR #123, opened 2020-06-21
Add create_topic and delete_topic

geoff2k:
This PR adds an Admin class that can be used to create and delete topics.  The API follows a pattern similar to the `Producer`, in that a handle is returned that may be `#wait`ed on by the caller for the operation to complete:

```ruby
admin = Rdkafka::Config.new(config).admin

create_topic_handle = admin.create_topic(topic_name, topic_partition_count, topic_replication_factor)
create_topic_handle.wait

delete_topic_handle = admin.delete_topic(topic_name)
delete_topic_handle.wait
```

Some implementation notes:

* I created a new `Admin` class to handle this functionality for a number of reasons: these methods are part of the ["Admin API"](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L5599) of the base C library, this gives us a place to hang other eventual administrative functions (ACLs anyone?), and TBH, from a symmetry perspective, I think create/delete should be in the same class.

* Even though the protocol (and [librdkafka](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L5906)) support creating multiple topics simultaneously, I've chosen to only create a single topic at a time as this is our (current) use case.

* The handle structure that I needed to track the create topic requests was similar enough to the existing `DeliveryHandle` class, that I pulled out an `AbstractHandle` class to contain the common bits.

* The `Bindings` module was getting quite long, so I split the callback functions out of it and put them in their own `Callbacks` module.  I left a comment in `callbacks_spec.rb` justifying the lack of unit tests for the module and offering some debugging advice.

* Test code coverage is at 💯 (per SimpleCov)

The previous summary for this PR can be [found here](https://gist.github.com/geoff2k/d8c4cf9e2617772505bc47b96389e0b1).


geoff2k:
Quick update now that CI has had a go at the code:

- one MRI build per batch has failed due to a segfault (first in [ruby 2.71](https://travis-ci.org/github/appsignal/rdkafka-ruby/builds/700633838) then in [ruby 2.5?](https://travis-ci.org/github/appsignal/rdkafka-ruby/builds/700655055))
- creating a topic seems to take longer in JRuby than in MRI (I extended the wait from 5s to 15s and Jruby is still unable to complete a CreateTopic successfully)
- in JRuby builds there are failures in some producer specs due to the handle REGISTRY not being cleared properly when a wait timeout is exceeded

I'll setup a JRuby environment and attempt to reproduce these locallly.

geoff2k:
JRuby failures were due to `:read` not being supported in the JRuby version of FFI (it was raising exceptions in the call backs).   Fixed in b9ba75a65e85f95cfd8ac30470d02863321a0e9a, and the build is now green. ✅

geoff2k:
@thijsc would you mind taking a look?

thijsc:
This looks like it's going in the right direction to me, encourage you to finish it up.

Adithya-copart:
@thijsc, Do you have any thoughts on adding a Admin API for creating/deleting topics versus defining them in the existing producer API?

I think defining these in producer API makes it easier to follow and also eliminate the need to establish a new socket connection and calling close after performing activities like creating or deleting a topic.

I like the approach of AbstractHandle by @geoff2k.

geoff2k:
> I think defining these in producer API makes it easier to follow and also eliminate the need to establish a new socket connection and calling close after performing activities like creating or deleting a topic.

@Adithya-copart I considered this, but eventually settled on putting them in a separate class (for reasons I've listed in the updated summary).  Also, topic creation **seems** like it would be a very rare occurrence?  Our intent is to use these methods in an engineer-facing, self-service, topic-maintenance interface, so the overhead of an extra socket creation per create/delete isn't that big of a deal for us.

How often are topics created and deleted in your environment?

geoff2k:
> This looks like it's going in the right direction to me, encourage you to finish it up.

@thijsc thanks for the encouragement!   I've  added the delete_topic method, tidied up best I can, and removed the "WIP" flag, so this is now ready for review.

The current build in CI is failing due to a [Segmentation fault](https://travis-ci.org/github/appsignal/rdkafka-ruby/jobs/705049099) which seems to sporadically occur in other builds that do not contain this code as well: [example1](https://travis-ci.org/github/appsignal/rdkafka-ruby/jobs/703628919), [example 2](https://travis-ci.org/github/appsignal/rdkafka-ruby/jobs/703626573).

Adithya-copart:
@geoff2k Initially, I thought that we could DRY the code by adding these methods to the producer directly and also eliminate the need to create a new socket connection. I agree with your logic and the code organization is better this way.

> How often are topics created and deleted in your environment?

Not often. We have an entire process for this and topics are created using the [scripts](https://github.com/apache/kafka/tree/trunk/bin) shipped with Kafka.

lib/rdkafka/admin.rb
```diff
@@ -0,0 +1,144 @@
...
+      # Create a rd_kafka_NewTopic_t representing the new topic
+      error_buffer = FFI::MemoryPointer.from_string(" " * 256)
+      new_topic_ptr = Rdkafka::Bindings.rd_kafka_NewTopic_new(
+        FFI::MemoryPointer.from_string(topic_name),
+        partition_count,
+        replication_factor,
+        error_buffer,
+        256
+      )
+      if new_topic_ptr == FFI::Pointer::NULL
```
Adithya-copart, line 50: We can use `new_topic_ptr.null?` if the return type is always a pointer,

geoff2k: @Adithya-copart thanks for noticing!  I have changed all similar comparisons to use `#null?` in fcd4a7b8a60071944b6bda486e7afa7619b3cc5a.

jlambert121:
This would be a great addition to allow us to manage topics with rdkafka-ruby as well as just using it for our producers and consumers.  Being a mostly ruby shop, the more things we can do in ruby the better.

thijsc:
Great work, d00 you think this is ready for final review and merge? I like that this is a separate module, makes sense to me.

Adithya-copart:
@thijsc This PR looks to be complete and is ready for a merge as far as I can tell.
I'll rebase #127 if you merge this first.

Issue #112, opened 2020-04-14
Consumer defaults encourage bad design


fritzblue:
While the producer defaults and examples correctly encourage batching out of the gate, the current consumer defaults (and all readily accessible examples) encourage a non-batch design. These defaults and examples should be updated to support the 99% use case, which is batching.

Encouraging handling messages one at a time, _even if the polling piece is batched under the hood_, in turn encourages secondary side effects like writing to a database, writing to another kafka topic, etc to be performed one at a time rather than in batches, which would be a bad design choice for users of this library.

I propose rethinking the defaults / examples of the consumer to be batch-based, like those of the producer.

thijsc:
I completely agree with you that batching is almost always preferred when consuming too.

We had some discussion about this in #99. I wrote up a batched example in this comment: https://github.com/appsignal/rdkafka-ruby/issues/99#issuecomment-553787214. I think it makes sense to put this example in the readme.

I still feel that the API and usage of enumerable, while not encouraging batching, do provide a very nice and Ruby-esque way of doing batching. Do you have some ideas of what a better API than this could be?

fritzblue:
Fortunately there's quite a bit of "prior art" on this subject we can refer to. Here is an example from the [official Java consumer](https://kafka.apache.org/25/javadoc/index.html?org/apache/kafka/clients/consumer/KafkaConsumer.html):

```java
     Properties props = new Properties();
     props.setProperty("bootstrap.servers", "localhost:9092");
     props.setProperty("group.id", "test");
     props.setProperty("enable.auto.commit", "false");
     props.setProperty("key.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
     props.setProperty("value.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
     KafkaConsumer<String, String> consumer = new KafkaConsumer<>(props);
     consumer.subscribe(Arrays.asList("foo", "bar"));
     final int minBatchSize = 200;
     List<ConsumerRecord<String, String>> buffer = new ArrayList<>();
     while (true) {
         ConsumerRecords<String, String> records = consumer.poll(Duration.ofMillis(100));
         for (ConsumerRecord<String, String> record : records) {
             buffer.add(record);
         }
         if (buffer.size() >= minBatchSize) {
             insertIntoDb(buffer);
             consumer.commitSync();
             buffer.clear();
         }
     }
```

[...]

PR #124, opened 2020-06-30
each_batch

mgrosso:
This PR provides `each_batch`, a method that allows an Rdkafka consumer to receive messages in batches while having control over the maximum size and maximum latencies of the batches.

The intent is to make it easier for users of the library to handle messages in bulk without blocking for unknown amounts of time. Consumers could implement that today using `poll`, since that's how this was implemented, but that implementation is error prone and would only rarely be specific to the business logic of the consumer.

The scope of this emerged in discussion with @thijsc  in the comments of https://github.com/appsignal/rdkafka-ruby/issues/112

### DIFF
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -471,5 +471,27 @@ module Rdkafka
     def closed_consumer_check(method)
       raise Rdkafka::ClosedConsumerError.new(method) if @native_kafka.nil?
     end
+
+    def each_batch(max_items: 100, max_latency_ms: 250, &block)
+      closed_consumer_check(__method__)
+      slice = []
+      loop do
+        start_time = Time.new.to_f
+        end_time = start_time + max_latency_ms / 1000.0
+        max_wait = end_time - Time.now.to_f
+        max_wait_ms = if max_wait <= 0
+                        0   
+                      else
+                        (max_wait * 1000).floor
+                      end 
+        message = poll max_wait_ms
+        slice << message if message
+        if slice.size == max_items || Time.now.to_f >= end_time
+          yield slice.dup
+          slice = []
+        end 
+        break if @closing
+      end 
+    end
   end
 end
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -489,7 +489,7 @@ module Rdkafka
         slice << message if message
         if slice.size == max_items || Time.now.to_f >= end_time
           yield slice.dup
-          slice = []
+          slice.clear
         end 
       end 
     end
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -487,7 +487,7 @@ module Rdkafka
                       end 
         message = poll max_wait_ms
         slice << message if message
-        if slice.size == max_items || Time.now.to_f >= end_time
+        if slice.size == max_items || Time.now.to_f >= end_time - 0.001
           yield slice.dup
           slice.clear
         end 
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -509,7 +509,7 @@ module Rdkafka
     #
     # @param max_items [Integer] Maximum size of the yielded array of messages
     #
-    # @param timeout_ms [Integer] Timeout of this poll
+    # @param timeout_ms [Integer] max time to wait for up to max_items
     #
     # @raise [RdkafkaError] When polling fails
     #
@@ -522,16 +522,15 @@ module Rdkafka
     def each_batch(max_items: 100, timeout_ms: 250, yield_on_error: false, &block)
       closed_consumer_check(__method__)
       slice = []
+      end_time = monotonic_now + timeout_ms / 1000.0
       loop do
         break if @closing
-        start_time = monotonic_now
-        end_time = start_time + timeout_ms / 1000.0
         max_wait = end_time - monotonic_now
         max_wait_ms = if max_wait <= 0
-                        0   
+                        0  # should not block, but may retrieve a message
                       else
                         (max_wait * 1000).floor
-                      end 
+                      end
         message = nil
         begin
           message = poll max_wait_ms
@@ -545,6 +544,7 @@ module Rdkafka
         if slice.size == max_items || monotonic_now >= end_time - 0.001
           yield slice.dup, nil
           slice.clear
+          end_time = monotonic_now + timeout_ms / 1000.0
         end 
       end 
     end
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -494,7 +494,18 @@ module Rdkafka
     # TopicPartitionList to {commit}.
     #
     # As with `each`, iteration will end when the consumer is closed.
-    # Exception behavior is also the same as with `each`.
+    #
+    # Exception behavior is more complicated than with `each`, in that if
+    # :yield_on_error is true, and an exception is raised during the
+    # poll, and messages have already been received, they will be yielded to
+    # the caller before the exception is allowed to propogate.
+    #
+    # If you are setting either auto.commit or auto.offset.store to false in
+    # the consumer configuration, then you should let yield_on_error keep its
+    # default value of false because you are gauranteed to see these messages
+    # again. However, if both auto.commit and auto.offset.store are set to
+    # true, you should set yield_on_error to true so you can process messages
+    # that you may or may not see again.
     #
     # @param max_items [Integer] Maximum size of the yielded array of messages
     #
@@ -502,10 +513,13 @@ module Rdkafka
     #
     # @raise [RdkafkaError] When polling fails
     #
-    # @yieldparam message [Batch] An array of received messages
+    # @yield [messages, pending_exception]
+    # @yieldparam messages [Array] An array of received Message
+    # @yieldparam pending_exception [Exception] normally nil, or an exception
+    # which will be propogated after processing of the partial batch is complete.
     #
     # @return [nil]
-    def each_batch(max_items: 100, timeout_ms: 250, &block)
+    def each_batch(max_items: 100, timeout_ms: 250, yield_on_error: false, &block)
       closed_consumer_check(__method__)
       slice = []
       loop do
@@ -522,18 +536,14 @@ module Rdkafka
         begin
           message = poll max_wait_ms
         rescue Rdkafka::RdkafkaError => error
-          # https://docs.confluent.io/2.0.0/clients/librdkafka/rdkafka_8h.html
-          #   RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS = 27,
-          if error.rdkafka_response == 27
-            yield slice.dup
-            return
-          else
-            raise
-          end
+          raise unless yield_on_error
+          raise if slice.empty?
+          yield slice.dup, error
+          raise
         end
         slice << message if message
         if slice.size == max_items || monotonic_now >= end_time - 0.001
-          yield slice.dup
+          yield slice.dup, nil
           slice.clear
         end 
       end 
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -518,7 +518,19 @@ module Rdkafka
                       else
                         (max_wait * 1000).floor
                       end 
-        message = poll max_wait_ms
+        message = nil
+        begin
+          message = poll max_wait_ms
+        rescue Rdkafka::RdkafkaError => error
+          # https://docs.confluent.io/2.0.0/clients/librdkafka/rdkafka_8h.html
+          #   RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS = 27,
+          if error.rdkafka_response == 27
+            yield slice.dup
+            return
+          else
+            raise
+          end
+        end
         slice << message if message
         if slice.size == max_items || Time.now.to_f >= end_time - 0.001
           yield slice.dup
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -472,13 +472,37 @@ module Rdkafka
       raise Rdkafka::ClosedConsumerError.new(method) if @native_kafka.nil?
     end
 
-    def each_batch(max_items: 100, max_latency_ms: 250, &block)
+    # Poll for new messages and yield them in batches.
+    #
+    # As with `each`, iteration will end when the consumer is closed.
+    # Exception behavior is also the same as with `each`.
+    #
+    # Rather than yield each message immediately as soon as it is received,
+    # each_batch will attempt to wait for as long as `timeout_ms` in order
+    # to create a batch of no more than `max_items` in size.
+    #
+    # Said differently, if more than `max_items` are available within
+    # `timeout_ms`, then `each_batch` will yield early with
+    # `max_items` in the array, but if `timeout_ms` passes by
+    # with fewer messages arriving, it will yield an array of fewer
+    # messages, possibly zero.
+    #
+    # @param max_items [Integer] Maximum size of the yielded array of messages
+    #
+    # @param timeout_ms [Integer] Timeout of this poll
+    #
+    # @raise [RdkafkaError] When polling fails
+    #
+    # @yieldparam message [Batch] An array of received messages
+    #
+    # @return [nil]
+    def each_batch(max_items: 100, timeout_ms: 250, &block)
       closed_consumer_check(__method__)
       slice = []
       loop do
         break if @closing
         start_time = Time.new.to_f
-        end_time = start_time + max_latency_ms / 1000.0
+        end_time = start_time + timeout_ms / 1000.0
         max_wait = end_time - Time.now.to_f
         max_wait_ms = if max_wait <= 0
                         0   
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -510,9 +510,9 @@ module Rdkafka
       slice = []
       loop do
         break if @closing
-        start_time = Time.new.to_f
+        start_time = monotonic_now
         end_time = start_time + timeout_ms / 1000.0
-        max_wait = end_time - Time.now.to_f
+        max_wait = end_time - monotonic_now
         max_wait_ms = if max_wait <= 0
                         0   
                       else
@@ -532,11 +532,17 @@ module Rdkafka
           end
         end
         slice << message if message
-        if slice.size == max_items || Time.now.to_f >= end_time - 0.001
+        if slice.size == max_items || monotonic_now >= end_time - 0.001
           yield slice.dup
           slice.clear
         end 
       end 
+    end
+
+    private
+    def monotonic_now
+      # needed because Time.now can go backwards
+      Process.clock_gettime(Process::CLOCK_MONOTONIC)
     end
   end
 end
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -472,20 +472,29 @@ module Rdkafka
       raise Rdkafka::ClosedConsumerError.new(method) if @native_kafka.nil?
     end
 
-    # Poll for new messages and yield them in batches.
-    #
-    # As with `each`, iteration will end when the consumer is closed.
-    # Exception behavior is also the same as with `each`.
+    # Poll for new messages and yield them in batches that may contain
+    # messages from more than one partition.
     #
     # Rather than yield each message immediately as soon as it is received,
     # each_batch will attempt to wait for as long as `timeout_ms` in order
-    # to create a batch of no more than `max_items` in size.
+    # to create a batch of up to but no more than `max_items` in size.
     #
     # Said differently, if more than `max_items` are available within
-    # `timeout_ms`, then `each_batch` will yield early with
-    # `max_items` in the array, but if `timeout_ms` passes by
-    # with fewer messages arriving, it will yield an array of fewer
-    # messages, possibly zero.
+    # `timeout_ms`, then `each_batch` will yield early with `max_items` in the
+    # array, but if `timeout_ms` passes by with fewer messages arriving, it
+    # will yield an array of fewer messages, quite possibly zero.
+    #
+    # In order to prevent wrongly auto committing many messages at once across
+    # possibly many partitions, callers must explicitly indicate which messages
+    # have been successfully processed as some consumed messages may not have
+    # been yielded yet. To do this, the caller should set
+    # `enable.auto.offset.store` to false and pass processed messages to
+    # {store_offset}. It is also possible, though more complex, to set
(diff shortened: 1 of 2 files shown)


spec/rdkafka/consumer_spec.rb
```diff
@@ -271,6 +271,14 @@ def send_one_message(val)
...
       consumer.subscribe("consume_test_topic")
+      100.times do |i|
+        report = producer.produce(
+          topic:     "consume_test_topic",
+          payload:   "payload #{i}",
+          key:       "key #{i}",
+          partition: 0
+        ).wait
+      end
       consumer.close
```
mgrosso, line 282: I played around with the tests a bit and realized that this one still passes even if you call close. To get it to fail with the `consumer.close` commented out I had to produce more than `50` messages.  I expected to have to produce at least one message to get it to fail in that state, but needing to produce 100 sort of surprised me, and tells me I need to think carefully about the tests I'm adding, given that they are all about latency and time outs.

lib/rdkafka/consumer.rb
```diff
@@ -441,5 +441,26 @@ def each
         end
       end
     end
+
```
mgrosso, line 444: I do intend to write a doc/comment

thijsc:
I like this direction, indeed close to what we discussed.

lib/rdkafka/consumer.rb
```diff
@@ -441,5 +441,26 @@ def each
...
+    def each_batch(max_items: 100, max_latency_ms: 250, &block)
+      slice = []
```
thijsc, line 446: Maybe there's a way to not have to create a new array?

mgrosso: > Maybe there's a way to not have to create a new array?

I believe it's essential for correct behavior to `dup` the array that we yield; otherwise, clients could accidentally muck with our array even as we're filling it back up, or create other unintended behaviours by accident. However I was able to convert the `slice = []` assignment on line 461 to `clear`, for what it's worth.

mgrosso:
@thijsc let me know if you want this as well in here:
```
M README.md
@@ -44,8 +44,10 @@ config = {
 consumer = Rdkafka::Config.new(config).consumer
 consumer.subscribe("ruby-test-topic")
 
-consumer.each do |message|
-  puts "Message received: #{message}"
+consumer.each_batch do |messages|
+  messages.each do |message|
+    puts "Message received: #{message}"
+  end
 end
 ```
 
I didn't want to presume.

**Comment by Adithya-copart** [comment]
@thijsc, @mgrosso I think this approach will fetch messages from different partitions as there are repeated calls to `poll` to create an array of messages referred to as a **batch**. The messages in a batch can be from different partitions as poll _could_ fetch messages from different partitions on each invocation. 

With "enable.auto.offset.store" set to true, all the messages in a batch could be marked as consumed even if the consumer fails while processing the first message in a batch. There could be a rebalance callback in between these polls which could mark the already fetched messages as processed. Users of this feature should be aware of these scenario's and handle offset commits themselves.

[Batches](https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#high-throughput) in librdkafka as well as ruby-kafka always imply a group of messages fetched from the same partition.

I think batch consuming is a useful feature to the library and adding the bindings to [rd_kafka_consume_batch](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L3343-L3373) might be the better approach. We can also leverage configuration options like `batch.num.messages` or `batch.size`.

**Comment by mensfeld** [comment]
> Batches in librdkafka as well as ruby-kafka always imply a group of messages fetched from the same partition.

And that's one of many reasons why Karafka is being switched to this library. While I can use the underlying API directly, I would be against mimicking this from ruby-kafka. I can understand the simplification, but with Ractors in Ruby 3.0, autofibers and other things, being able to distribute the workload based on multi-partition (or multitopic) batches seems more natural. It is more complex to grasp but it provides much bigger flexibility.

**Comment by Adithya-copart** [comment]
@mensfeld I could not find a definition for batches on the consumer side in librdkafka. I retract the statement that `ruby-kafka` and librdkafka share the same definition for consumer batches.

The legacy `batch_consume` API is per partition which led me to believe it shares the definition with `ruby-kafka`. The `rd_kafka_consume_batch_queue` is to fetch messages from the internal queue which is a collection of messages from multiple topics/partitions.

I don't see a batch consume API for the high-level consumer currently used. An alternative is suggested in this issue: https://github.com/edenhill/librdkafka/issues/928

> distribute the workload based on multi-partition (or multitopic) batches seems more natural

Did you elaborate on this anywhere?
I'm interested to learn more as a JRuby user.

**Comment by Adithya-copart** [comment]
In the current approach, auto-commit configuration property cannot be combined with `each_batch`. Users need to elect one of the following behaviors and implement it when an exception is encountered during processing:

1. Rollback the entire batch by skipping the offset commits.
2. Partially commit the already processed offsets.

**Comment by mgrosso** [comment]
@Adithya-copart based on your comments, is it fair to say you're recommending two changes?:

1- raise an exception if `each_batch` is called on a consumer where `enable.auto.commit` is true
2- document more clearly that the batches will contain messages from a mix of partitions and that users are responsible for storing offsets on all of those partitions (ie: calling `store_offsets` with just the last message might lead to sadness), and that this responsibility for committing offsets includes deciding how to deal with errors arising from what may well be a mixed partition batch.

**Comment by Adithya-copart** [comment]
@mgrosso Yeah. Any thoughts on handling a rebalance callback that revokes assigned partitions while we accumulate a batch?

I think the chances of this happening are low but we might still need to consider it and try to handle it in the library. We can do this by storing the offsets before creating a batch and commit them when a rebalance happens. We will end up discarding the accumulated messages.

Otherwise, we can stop polling upon rebalance and yield the batch accumulated until that point so that client can commit the offsets after processing. We have to synchronize between the rebalance callbacks and the batch polling to complete the partition unassign.

**Comment by Adithya-copart** [comment]
> 	/** Group rebalance in progress */
> 	RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS = 27,

If this is raised when we call poll, we can probably rescue the error and stop accumulating the batches at that point and yield the batch.

**Comment by mensfeld** [comment]
> distribute the workload based on multi-partition (or multitopic) batches seems more natural

Did you elaborate on this anywhere?
I'm interested to learn more as a JRuby user.

It's not the place ;) we can DM (maciej@mensfeld.pl) - just drop me a line.

**Comment by mgrosso** [comment]
> > /** Group rebalance in progress */
> > RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS = 27,
> 
> If this is raised when we call poll, we can probably rescue the error and stop accumulating the batches at that point and yield the batch.

Hi @Adithya-copart ,

I've enhanced the docs to describe why it's important to store_offsets and to be clear that batches yielded will contain a mix of partiions.

It looks like raising an exception if `each_batch` is called with particular consumer configurations would require larger changes than I think we want here. If I'm wrong let me know how you would do it, but it seems to me that the configuration hash is no longer accessible when `each_batch` is called and that storing it would change the very minimal character of `rdkafka-ruby`.

Regarding `RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS`, I could use some more explanation, because I'm not clear on a couple things. The case of `each_batch` with manual `store_offset` seems to have the same issues as using manual `store_offset` in the case of `poll` or an `each`/Enumerable derived `take(3)` with manual commits.  In all of these cases, when the "rebalance in progress" exception is raised there may be some messages that are buffered and unprocessed, and others that are processed but still not committed. If the caller tries to store_offset and commit after the exception wouldn't it be at great risk of not actually changing the first offset seen by the new owner of the partition because of race conditions?  In the specific case of `each_batch`, how does it help the calling code to catch the exception, yield outstanding buffered messages, and then reraise the exception, given that another consumer is gauranteed to receive those messages?  and how should that "catch-yield-raise" code handle any exceptions that come from the `yield` while it's hanging onto a "rebalance in progress" exception that still needs to be re-raised?   I just want to make sure the benefit of the additional complexity outweighs it's risks.

Thanks,
Matt

**Comment by mgrosso** [comment]
> > > /** Group rebalance in progress */
> > > RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS = 27,
> > 
> > 
> > If this is raised when we call poll, we can probably rescue the error and stop accumulating the batches at that point and yield the batch.
> 
> 
> Regarding `RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS`, I could use some more explanation, because I'm not clear on a couple things. ...

ok, @Adithya-copart , I believe I'm beginning to understand with respect to the rebalance in progress exception.  Will consumers that *don't* lose their partitions subsequently receive messages with the next unconsumed offset? or reset to the next uncommitted offset?

If it's the former, then I definitely understand why your suggested catch/yield behavior is mandatory to avoid a gap of unprocessed messages that had been buffered when the exception propagated.  If it's the latter, then the yield would be unnecessary and lead to duplicate processing. Most consumers are probably "at-least-once" anyhow but still.
 
Now suppose a consumer with a single partition simply allowed the exception to propagate, as many would, such that the brokers never hear from this consumer again and it's recreated by some process manager, as is also typical. This would result in another partition reallocation. Hopefully most of those occasions don't cause additional rebalance in progress exceptions in other consumers in the group as a chain reaction would be unfortunate.

What is the recommended behavior for consumer receiving that particular exception?  sleep briefly in the rescue and retry?

Thanks for the feedback earlier,
Matt

**Comment by Adithya-copart** [comment]
@mgrosso I'm not sure of how things work with auto offsets turned off. The specs use the default configuration property of `enable.auto.commit`.

I found this overall workflow diagram in [librdkafka](https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md) which tells us that a consumer will wait for the offsets to be committed before revoking the assigned partitions as part of the rebalance.

I will try to test the scenario where a rebalance is triggered while a batch of messages are being accumulated. I don't know if `rd_kafka_consumer_poll` raises `RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS` or returns a null pointer.

- If an error is raised, we need to rescue, stop the accumulation, let the consumer process the batch and commit the offsets which will trigger the revoke of the partitions as per the flow diagram.

- If an error is not raised and a null pointer is returned, we'll hit the timeout and yield the batch for processing anyway. The partitions will be revoked when the consumer commits the offsets. I wonder if this means that a consumer can only manually commit the offsets after processing the entire batch.

> Will consumers that don't lose their partitions subsequently receive messages with the next unconsumed offset? or reset to the next uncommitted offset?

I'm not sure if I follow this correctly but this is my understanding: A consumer will stop consuming messages and mark the last consumed message in a partition in the final offset commit provided there are no processing errors. So, the next offset in the partition based on the last committed offset will be consumed after rebalance in the same or a different consumer.

image

**Comment by mgrosso** [comment]
> `RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS`

@Adithya-copart thanks, that diagram was helpful. I've implemented the behavior and pushed a commit. let me know what you think.

**Comment by Adithya-copart** [comment]
@mgrosso The changes look good to me.

I tried running consumers with `enable.auto.commit` set to false. `RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS` is not raised during `poll` in my local setup with `docker-compose` used in the test suite.

This [script](https://gist.github.com/Adithya-copart/ffc4c4ce590d20f40701e8110abd6e54) in JRuby results in some messages being processed more than once with `enable.auto.commit` set to false. I'm trying to see if I can find the reason for this. The messages are not consumed more than once with `enable.auto.commit:true` and `auto.offset.reset: earliest`.

**Comment by Adithya-copart** [comment]
[Updated JRuby script that uses manual commits](https://gist.github.com/Adithya-copart/3c598b696a1a6479cdfb35d791c49949).

I finally had some success testing this workflow with `enable.auto.commit: false`. All produced messages are processed exactly once with consumers rejoining the group multiple times. We could test some failure scenarios by triggering random exceptions when a batch is being processed that results in the consumer exiting the group and rejoining.

The pain-points I noticed so far are:

1. Lack of information about the offsets in the batch for manual commits.
2. Lack of a hook in the `Opaque` class that can be used by consumers to commit the offsets during rebalance.

**Information about offsets in the batch:**

We need this so that clients can avoid iterating the messages in a batch for the latest offset processed in each partition to commit manually. Since we accumulate to build a batch, we can store the offsets in a `TopicPartitionList` or a `Hash` and expose it to the clients.

**Rebalance hook:**

Consumers should commit their offsets before partitions are revoked to avoid re-processing the same messages. We don't have a hook for this. We have `call_on_partitions_revoked` that is called after the partitions are revoked so we can add a similar hook that will be called before revoking the partitions.

The script does the following:

- Update offsets of the consumer group to make sure old messages are not processed.
- Produce 100_000 messages to a topic with multiple partitions.
- Spawn 't' threads with 1 consumer in every thread. Each consumer will restart 'n' times after processing 10 batches. This triggers a consumer rebalance.
- Validate the messages processed.

**Comment by mgrosso** [comment]
> We need this so that clients can avoid iterating the messages in a batch for the latest offset processed in each partition to commit manually. Since we accumulate to build a batch, we can store the offsets in a `TopicPartitionList` or a `Hash` and expose it to the clients.

@Adithya-copart thank you, this is awesome testing. I have an observation and 3 questions.

seems like there are a few different valid use cases with respect to offsets.

1. auto everything should be at least once.
2. auto commit on with auto.offset.store=false and calls to `store_offset(message)` should be at least once.
3. same as 2, but with auto commit off and manual `commit(nil,...)` should also be at least once.
4. auto commit off with manual partition/offset tracking and manual `commit(list,...)` could be used to implement either at most once or at least once depending on whether you did the processing before or after the commit.

questions...

1) I'd be willing to implement a more "integration" test style of test validating results in the style you suggested but I'll look for guidance first on whether that should be a separate rake task instead of an rspec and what sorts of runtimes would lead us to choose one versus another.  What sorts of choices for numbers and threads and events would be sufficient to test this feature? Would you fail the test on duplicate processing in the "at-least-once" scenarios above? which of those usage types (1..4) above would you want to see in that?

2) As a side note, does the convenience `store_offsets` method as implemented today need to work around the issue  https://github.com/edenhill/librdkafka/issues/1811 you highlighted in that code?  I wouldn't do that in this PR, but I'm curious. I also wonder if users are already implementing that workaround, such that any fix would cause skipped messages :-\.  Perhaps a line of expectation setting in the docs?

3) To address the "Information about offsets in batch" issue you identify, what do you think about offering two convenience methods: `store_batch_offsets(batch)` that accepts the batch and stores the correct offset for each partition, and also a `commit_batch_offsets` that similarly builds the list and commits it?  That avoids complexities in the issue of storing state for a callback that might not be used, yielding an additional parameter that also might not get used, or yielding something with behavior.

Thanks

**Comment by Adithya-copart** [comment]
@mgrosso It seems to me that approaches 2 and 3 may not be used for batch processing and the existing specs might already cover the `each` style message processing.

1. Personally, I don't think we need to add integration style of tests for all scenario's(1..4) but we could definitely have tests for at least once guarantee in `each_batch` for the "auto everything" case and the last one with manual commits. I went with arbitrary numbers in the test script and I think that specs should be fine.

2. It looks like `rd_kafka_offset_store` which is internally used by `store_offset(message)` [increments the offset](https://github.com/edenhill/librdkafka/blob/b3223ecea638bfacc6d9ec981aa4df1d22ee9542/src/rdkafka_offset.c#L663-L664) by 1.

3. I agree that an additional parameter of partition specific offsets may not be used by the consumers every time. I'm against `store_batch_offsets(batch)` if the idea is to iterate over the batch upon invocation of this method. It might be easier to store the offset in a ivar while accumulating the batch and re-use it in something like `commit_batch_offsets` even if we chose to not yield it as an additional parameter. 

@thijsc Any thoughts or comments?

File: lib/rdkafka/consumer.rb [hunk 4]
```diff
@@ -441,5 +441,71 @@ def each
...
+                      else
+                        (max_wait * 1000).floor
+                      end 
+        message = nil
+        begin
+          message = poll max_wait_ms
+        rescue Rdkafka::RdkafkaError => error
+          # https://docs.confluent.io/2.0.0/clients/librdkafka/rdkafka_8h.html
+          #   RD_KAFKA_RESP_ERR_REBALANCE_IN_PROGRESS = 27,
+          if error.rdkafka_response == 27
```
Adithya-copart, line 496: @mgrosso Theoretically, we could receive any error during the calls to `poll`.
What do you think about yielding the batch followed by raising the received error?

If we do not do this, we might fail the at least once test in "auto everything" scenario.

mgrosso: Hi @Adithya-copart , I'm happy to do that but two things about it trouble me:

1. Tell me if I'm missing something here but I don't think we can offer *any* guarantee to the "auto-everything" consumer and it should be discouraged for any use case that needs one.

My reasoning is that the client code processing of the message can take an arbitrary amount of time before raising an uncaught exception, being zapped by the kernel oom handler, or losing power to the node.

At that random point when message processing fails for the "auto everything" consumer the asynchronous offset management may or may not have stored the offset or committed it to the broker.  If it *has*, then there will be an unprocessed message.

However in the opposite case, suppose the processing code successfully handled one or more messages but they haven't been committed yet. When another consumer is assigned this partition they will duplicate the processing.

I see this issue as fundamental to "auto everything" rather than unique to `each_batch`. ie, if we do it here, we should also do it in `each`.  I can see how yielding the batch for "normal exceptions" helps the "auto-everything" consumer get closer to "almost-always-once" so I think it makes sense for rebalance. 

2. My second concern is that passing the outstanding buffered messages in many if not most of the other error possibilities will just guarantee duplicate processing for consumers who manage their own commits or offsets.  Many of those errors names suggest that any subsequent attempt to commit will fail. eg: RD_KAFKA_RESP_ERR_KAFKA_STORAGE_ERROR
https://github.com/edenhill/librdkafka/blob/39796d359898c07ea422849e6d7cd34cd13ec466/src/rdkafka.h#L260
Of course that's not horrible for "at-least-once" in and of itself, but it doesn't seem especially helpful to me either. 

Do those issues make sense?  Do you think it's worth it instead to identify a subset of exceptions worth yielding in both `each` and `each_batch`? that seems daunting to me, and like it might require maintenance over time.

Thanks again for all your careful help with this PR.

Adithya-copart: > My reasoning is that the client code processing of the message can take an arbitrary amount of time before raising an uncaught exception, being zapped by the kernel oom handler, or losing power to the node.
> 
> At that random point when message processing fails for the "auto everything" consumer the asynchronous offset management may or may not have stored the offset or committed it to the broker. If it has, then there will be an unprocessed message.

As long as the library delivers the message to the client, we should be upholding the at least once delivery guarantee.
If the client fails to process it, the responsibility rests with the client but not the library.

I do agree that the actual exception while accumulating the buffer will be obscured if it is followed by another exception in the client code if we yield it. If it happens in client space, clients should be handling it.

For clients not using auto-commit, if a consumer processing a message fails to commit the offset, the client can choose to  undo the processing or re-process the message again on a subsequent delivery. Other client space exceptions can be handled as per the clients choosing.

> However in the opposite case, suppose the processing code successfully handled one or more messages but they haven't been committed yet. When another consumer is assigned this partition they will duplicate the processing.

I think duplicate processing is okay if the at least once guarantee is upheld by `rdkafka`. Consumers that need at-most once can always use manual commits by ensuring the processing of the message/batch and committing the offsets is atomic. This would prevent any duplicate processing.

> Do you think it's worth it instead to identify a subset of exceptions worth yielding in both each and each_batch?

I don't think `rdkafka` needs to change anything in the `each` case. We call `poll` whose result could be either a message or an exception. The client receives it directly and there is no buffer involved.

In the case of `each_batch`, we are introducing a buffer which is not available in the client space until it is yielded. So, if we do not yield this buffer in case of an exception, the client will never receive it but could be marked as processed which is the case currently if we other `Enumerable` methods other than `each`.

Adithya-copart: > Thanks again for all your careful help with this PR.

Thanks for your kind words Matt. Glad I could be of help.
I'm learning quite a bit from you in return as well.

thijsc: Interesting discussion!

I do think the offsets should in principle only be added if the batch was actually completed. Maybe we should disable automatic offset storage by default and set the offsets to the one of the latest message after the `yield`?

spec/rdkafka/consumer_spec.rb
tkornai: To make the test fail with consumer.close commented out there is no need to produce new messages, it is enought to raise the poll timeout, e.g.:
expect(consumer.poll(10000)).to be_nil

lib/rdkafka/consumer.rb
tkornai: Hi @thijsc, when proposing to set `enable.auto.offset.store = false` by default, what scope do you mean? How could we keep the functionality intact for existing users of the library?

thijsc: > When proposing to set enable.auto.offset.store = false by default, what scope do you mean?

The default config is set [here](https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/config.rb#L57).

> How could we keep the functionality intact for existing users of the library?

I could see us also calling `store_offset` in `each`. Then it would set the offset after a successful yield, which is more robust anyway. The only thing that will then be affected is people using `poll` manually. There are probably not many, if any at all out there. We could add a warning in the release notes and document it.

lib/rdkafka/consumer.rb
```diff
@@ -441,5 +441,71 @@ def each
...
+    # @raise [RdkafkaError] When polling fails
+    #
+    # @yieldparam message [Batch] An array of received messages
+    #
+    # @return [nil]
+    def each_batch(max_items: 100, timeout_ms: 250, &block)
+      slice = []
+      loop do
+        break if @closing
+        start_time = Time.new.to_f
```
mensfeld, line 482: can we use monotonic clock here please?

thijsc:

thijsc:
Any chance you could rebase?

mgrosso:
> Any chance you could rebase?

Hi @thijsc , absolutely. I'll have time tomorrow. sorry I didn't see your earlier prompt. 

Also, I'm hoping this commit might address the remaining concern about pending unhandled messages when an exception arises:  https://github.com/mgrosso/rdkafka-ruby/commit/a7f2784

mgrosso:
rewriting the spec that failed to remove timing issues.

mgrosso:
@thijsc I believe I've addressed all outstanding comments at the point.  Thanks for your patience, since I stepped away from this for a few months.

Those that want to keep `enable.auto.commit` and `enable.auto.offset.store` both true can set `yield_on_error` to true and handle events that would otherwise be lost, although that configuration remains inherently risky. The extra yielded parameter is the exception that will be thrown when that batch is complete. This complexity is avoided by default because yield_on_error defaults to false. I think this should address @Adithya-copart 's comments.

Time comparisons use monotonically increasing time now, thanks to a comment whose author I can't find right now.

Let me know what else could be done to improve the PR. Also, Happy New Year!

PR #108, opened 2020-02-04
Support JRuby and fix travis failures

Adithya-copart:
Re-opening #106 here to eliminate the noise due to numerous debugging commits.
I find it easier to trigger travis builds than running the specs locally.

The changes in this PR also include @gaffneyc's changes in #107:

1. Workarounds for the lack of `read_int64`, `read(:pointer)`, `read_string_to_null` in JRuby.
2. Avoid using FFI::AutoPointer objects inside `Rdkafka::Consumer` and explicitly calling `rdkafka_destroy` during close.
3. Avoid calls to `rd_kafka_consumer_poll` after consumer is closed.
4. Use `RbConfig::CONFIG` to check the host OS which is JRuby friendly.
5. Switch to [confluentinc/cp-kafka](https://hub.docker.com/r/confluentinc/cp-kafka) and [confluentinc/cp-zookeeper](https://hub.docker.com/r/confluentinc/cp-zookeeper) licensed with a Apache 2.0 license.
6. Add JRuby to the list of rubies in `.travis.yml`.
7. Remove the usage of `AutoPointer` as it'll lead to closing the socket connections when a child process created using `Kernel#fork` exits which could lead to segmentation faults.

**Findings:**

`E, [2020-01-29T06:17:58.952506 #9416] ERROR -- : rdkafka: [thrd:GroupCoordinator]: 1/1 brokers are down` in specs is trivial as per https://github.com/confluentinc/confluent-kafka-dotnet/issues/1129#issuecomment-561516728 

I'll gladly rebase this PR if you'd wish to merge #107 first.

PR #125, opened 2020-07-05
Don't enable SimpleCov in CI

geoff2k:
While working on #123, a number of my builds segfaulted and failed during their runs under Travis ([example 1](https://travis-ci.org/github/appsignal/rdkafka-ruby/jobs/705049099)).  I investigated and noticed that segfaults appear to occur sporadically in other builds that did **not** contain my code: [example2](https://travis-ci.org/github/appsignal/rdkafka-ruby/jobs/703628919), [example 3](https://travis-ci.org/github/appsignal/rdkafka-ruby/jobs/703626573).

I have been unable to reproduce the segfaults locally under any of the versions of Ruby we test with. 😞 

Per google searches, [SimpleCov](https://github.com/colszowka/simplecov) seems to be anecdotally related to segmentation faults experienced running builds in CI, so this PR checks for a well-known environment variable `CI` (supported under both [Travis](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) and [CircleCI](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables)) and doesn't enable SimpleCov if present.

If this does not solve the issue, we may need to use something like [travis-codedump](https://github.com/springmeyer/travis-coredump) to diagnose the segfaults.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,8 @@
-require "simplecov"
-SimpleCov.start do
-  add_filter "/spec/"
+unless ENV["CI"] == "true"
+  require "simplecov"
+  SimpleCov.start do
+    add_filter "/spec/"
+  end
 end
 
 require "pry"


mensfeld:
I have a similar problem - but that is not because of simplecov. That's due to some issues with FFI and Rubys GC. There were some bypasses suggested (including by me) but they don't seem to work in all the scenarios.

geoff2k:
@mensfeld In your estimation, are these worth trying to track down?  In other words, while the failures in CI are annoying, I'd be more willing to ignore them if we knew they didn't point to some problem with the code under test, if that makes sense.

Adithya-copart:
@geoff2k I ran into similar issues while open sockets are closed during the shutdown sequence.

See https://github.com/appsignal/rdkafka-ruby/pull/108#issuecomment-586742532 and [CI run](https://travis-ci.org/github/appsignal/rdkafka-ruby/builds/651150747). 

I ended up calling `close` in the specs where sockets were left open after some debugging at that time.

mensfeld:
> @mensfeld In your estimation, are these worth trying to track down? In other words, while the failures in CI are annoying, I'd be more willing to ignore them if we knew they didn't point to some problem with the code under test, if that makes sense.

The question here is: will it occur in production. And the answer is: I don't know. I took several countermeasures to make sure it won't happen in production (in Waterdrop). I tackled majority of cases in specs as well but there's still something off.

I've been running Waterdrop 2.0 (with rdkafka) in scale of up to 10k rq/s with auto-scaling on aws and things that could potentially trigger this as well as with instrumentation to detect something like that upon process exit and for the past several months it never occured anywhere beyond CI.

However, I still see this as an issue worth investigating.

thijsc:
> I've been running Waterdrop 2.0 (with rdkafka) in scale of up to 10k rq/s with auto-scaling on aws and things that could potentially trigger this as well as with instrumentation to detect something like that upon process exit and for the past several months it never occured anywhere beyond CI.
> However, I still see this as an issue worth investigating.

Same for us, we run this in production and have never once seen it happen. Still unsure why the spec context matters so much. Probably because in production you never close and recreate consumers/producers.

PR #126, opened 2020-07-09
Raise an exception if @native_kafka is nil

geoff2k:
## The problem

Many methods in the `Producer` and `Consumer` classes are dependent on being able to pass the value of `@native_kafka` instance variable to underlying librdkafka bindings.

This can lead to trouble, however if the `#close` method has been called (which sets `@native_kafka` to nil), and can lead to unrecoverable segfaults.

## The proposed solution

Do a pre-flight check in every method that is dependent on `@native_kafka` being non-nil so that we can raise a proper Ruby exception to the caller instead of facing them with a 1000+ C line backtrace and crash log ([example](https://gist.github.com/geoff2k/d95d54185d4ba556368da4c049ea4026)).

In my opinion, we should strive for gems that *never* allow a ruby developer to see a C stack trace, or more glibly, isnt this:

```
     Rdkafka::Consumer::ClosedConsumerError:
       Illegal call to subscribe on a closed consumer
```

better than this? 😉 

```
-- Control frame information -----------------------------------------------
c:0046 p:---- s:0246 e:000245 CFUNC  :rd_kafka_topic_partition_list_add
c:0045 p:0022 s:0239 e:000238 BLOCK  /Users/geoff2k/src/kafka/rdkafka-ruby/lib/rdkafka/consumer.rb:45 [FINISH]
c:0044 p:---- s:0235 e:000234 CFUNC  :each
c:0043 p:0033 s:0231 e:000230 METHOD /Users/geoff2k/src/kafka/rdkafka-ruby/lib/rdkafka/consumer.rb:44 [FINISH]
....
```

### DIFF
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -36,6 +36,8 @@ module Rdkafka
     #
     # @return [nil]
     def subscribe(*topics)
+      raise "Illegal call to #subscribe after closing the consumer" if @native_kafka.nil?
+
       # Create topic partition list with topics and no partition set
       tpl = Rdkafka::Bindings.rd_kafka_topic_partition_list_new(topics.length)
 
@@ -58,6 +60,8 @@ module Rdkafka
     #
     # @return [nil]
     def unsubscribe
+      raise "Illegal call to #unsubscribe after closing the consumer" if @native_kafka.nil?
+
       response = Rdkafka::Bindings.rd_kafka_unsubscribe(@native_kafka)
       if response != 0
         raise Rdkafka::RdkafkaError.new(response)
@@ -72,6 +76,8 @@ module Rdkafka
     #
     # @return [nil]
     def pause(list)
+      raise "Illegal call to #pause after closing the consumer" if @native_kafka.nil?
+
       unless list.is_a?(TopicPartitionList)
         raise TypeError.new("list has to be a TopicPartitionList")
       end
@@ -98,6 +104,8 @@ module Rdkafka
     #
     # @return [nil]
     def resume(list)
+      raise "Illegal call to #resume after closing the consumer" if @native_kafka.nil?
+
       unless list.is_a?(TopicPartitionList)
         raise TypeError.new("list has to be a TopicPartitionList")
       end
@@ -120,6 +128,8 @@ module Rdkafka
     #
     # @return [TopicPartitionList]
     def subscription
+      raise "Illegal call to #subscription after closing the consumer" if @native_kafka.nil?
+
       ptr = FFI::MemoryPointer.new(:pointer)
       response = Rdkafka::Bindings.rd_kafka_subscription(@native_kafka, ptr)
 
@@ -142,6 +152,8 @@ module Rdkafka
     #
     # @raise [RdkafkaError] When assigning fails
     def assign(list)
+      raise "Illegal call to #assign after closing the consumer" if @native_kafka.nil?
+
       unless list.is_a?(TopicPartitionList)
         raise TypeError.new("list has to be a TopicPartitionList")
       end
@@ -164,6 +176,8 @@ module Rdkafka
     #
     # @return [TopicPartitionList]
     def assignment
+      raise "Illegal call to #assignment after closing the consumer" if @native_kafka.nil?
+
       ptr = FFI::MemoryPointer.new(:pointer)
       response = Rdkafka::Bindings.rd_kafka_assignment(@native_kafka, ptr)
       if response != 0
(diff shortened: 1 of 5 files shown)


lib/rdkafka/consumer.rb
```diff
@@ -393,7 +419,7 @@ def commit(list=nil, async=false)
     #
     # @return [Message, nil] A message or nil if there was no new message within the timeout
     def poll(timeout_ms)
-      return unless @native_kafka
+      closed_consumer_check(__method__)
```
geoff2k, line 422: This is a change in behaviour, but I'm curious if we really want to allow a `#poll` when the underlying kafka instance is closed?

Adithya-copart:
Nice. I run into these faults while pasting code in a `bundle console` session too. The current approach in this PR is definitely a improvement that what we currently have. Personally, I think it would be nice if we could:

- Subclass the error from `RdkafkaError` so that clients can rescue `RdkafkaError` or `RdkafkaClosedConsumerError` based on their preference.
- Move `@native_kafka` nil check to the native function bindings and raise an error there directly.

It might be possible to type check after reading ffi documentation and a couple of blog posts. I came across https://www.varvet.com/blog/advanced-topics-in-ruby-ffi/ with some nice tricks that we could use.

I will try to test some approaches suggested there for type checks using `FFI::DataConverter`. The current approach using `closed_consumer_check` in multiple places feels a bit like boilerplate to me.

Adithya-copart:
@geoff2k Pushed https://github.com/appsignal/rdkafka-ruby/pull/127 using `FFI::DataConverter` that does a nil-check when a C-function is being called with `@native_kafka`.

I think we could use this pattern in multiple places so that `bindings.rb` can directly tell us what is being passed to a C-function in places like: https://github.com/appsignal/rdkafka-ruby/blob/16a4497f59f4d369ebfb4ad0784161400fd96424/lib/rdkafka/bindings.rb#L174

geoff2k:
> Subclass the error from RdkafkaError so that clients can rescue RdkafkaError or RdkafkaClosedConsumerError based on their preference.

I like this idea in concept, but RdkafkaError does not lend itself to general subclassing because it is meant to be based on a numeric return code from the underlying librdkafka library (i.e. one of the values of [`rd_kafka_resp_err_t`](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L248)), and we don't have one in this case, since this is more of a "you are using the gem wrong" type of error.

> Move @native_kafka nil check to the native function bindings and raise an error there directly.

While I don't disagree that we should add these types of checks to the bindings (engineers working on the gem itself should have some layer of protection against this), one of the things that I like about the current messages is that they succinct and they are at the right level of abstraction -- what we're trying to do with these errors is point out that the calling sequence for the gem's API has been violated, and that users of the gem should not do this.

I worry about moving these check "too low" in the stack, I'm trying to avoid an average gem users who callls `#subscribe` on a previously closed consumer receiving a message about null pointers in a call to `rd_kafka_topic_partition_list_add`, as it would  not be clear to them what to do with this information.

Does that make sense?

Adithya-copart:
>  RdkafkaError does not lend itself to general subclassing because it is meant to be based on a numeric return code

Do you think we can change `RdkafkaError` to accommodate this use-case?

> I like about the current messages is that they succinct and they are at the right level of abstraction -- what we're trying to do with these errors is point out that the calling sequence for the gem's API has been violated, and that users of the gem should not do this.

We can still add a [custom error](https://github.com/appsignal/rdkafka-ruby/pull/127/files#diff-e2c7aea3be322c3c525fdf482a471043R50) that helps users understand that a consumer/producer is closed rather than a null pointer error. We'll lose the method name in the error message though.

However, one of the important change you did in #126 is a separate error for the producer and consumer classes. This should be possible in #127 with slight modifications too. This abstraction will help users write logic like:

```
begin
  consumer.each |message|
       # do something
       producer.produce(...)
  end
rescue Rdkafka::ClosedConsumerError
  consumer = config.consumer
  retry
rescue Rdkafka::ClosedProducerError
  producer = config.producer
  retry
end
```

lib/rdkafka/consumer.rb
thijsc: I think raising an exception is fine in that case.

thijsc:
I'm ok with adding a few more error types. Could you do that and rebase? I think this is ready to merge then.

geoff2k:
@thijsc @Adithya-copart apologies for the latency on getting back to this.   I've made the following updates as discussed:
* rebased to current `main`
* created a base error class and two new subclasses in error.rb : https://github.com/appsignal/rdkafka-ruby/pull/126/files#diff-c7c65a1cf345f93492d2140edbb0369d

Hopefully this is now good to 🚢 .

thijsc (approved):

PR #127, opened 2020-07-11
Use FFI::DataConverter to define native handle and add type/nil checks

Adithya-copart:
Alternate take on #126

### DIFF
--- a/lib/rdkafka/error.rb
+++ b/lib/rdkafka/error.rb
@@ -1,6 +1,9 @@
 module Rdkafka
+  # Base error class.
+  class BaseError < RuntimeError; end
+
   # Error returned by the underlying rdkafka library.
-  class RdkafkaError < RuntimeError
+  class RdkafkaError < BaseError
     # The underlying raw error response
     # @return [Integer]
     attr_reader :rdkafka_response, :message_prefix
@@ -8,6 +11,7 @@ module Rdkafka
     # @private
     def initialize(response, message_prefix=nil)
       raise TypeError.new("Response has to be an integer") unless response.is_a? Integer
+
       @rdkafka_response = response
       @message_prefix = message_prefix
     end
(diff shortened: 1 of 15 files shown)


Adithya-copart:
Added 26beb479dcd87a570a248cd307cdd8633694f6b5 as there are two calls to `to_native` as mentioned in https://github.com/ffi/ffi/issues/807

thijsc:
Looking good, what's still draft about it?

Adithya-copart:
Hey @thijsc ,

I'm thinking to add more classes that use `FFI::DataConverter` so that references like `:pointer, :int32 etc` can be replaced with these classes. It would be a refactor like the `Handle` class. What are your thoughts?

Should we rather merge the `Handle` class first?

Also, this PR has a `Rdkafka::BaseError` that could be used as the base class of all errors that can be raised in the library. I inherited the new errors introduced in this PR from this class. Is this an approach you're fine with?

spec/rdkafka/producer_spec.rb
```diff
@@ -407,4 +407,23 @@
...
+
+  it "should close open producer sockets before they are GC'd" do
+    skip if defined?(JRUBY_VERSION)
+
+    @cleared = false
+    producer = rdkafka_config.producer
+    ObjectSpace.define_finalizer(producer, finalizer_proc)
+    producer = nil
+    # Request GC run.
+    RUBY_PLATFORM.casecmp?('java') ? java.lang.System.gc : GC.start
```
Adithya-copart, line 425: GC run is not guaranteed upon invocation as per the docs, so this test could be unreliable as-is.

The approach to validate the instance variable is not working in JRuby even though the finalizer is called.
https://github.com/jruby/jruby/issues/6353

thijsc:
Let's merge the handle class, and rebase this then.

lib/rdkafka/error.rb
```diff
@@ -57,4 +61,13 @@ def initialize(response, topic_partition_list, message_prefix=nil)
       @topic_partition_list = topic_partition_list
     end
   end
+
+  # Consumer closed error.
+  class ConsumerClosedError < BaseError; end
```
Adithya-copart, line 66: Relevant specs are needed.

thijsc:
Could you rebase this one?

thijsc:

thijsc:
Could you rebase this?

thijsc:
@Adithya-copart Any chance you could rebase this one?

thijsc:
Closing this one for the moment to get to a clean slate, please reopen when it can be merged :-)

PR #128, opened 2020-07-12
Fix a couple of issues in the Metadata class

geoff2k:
## The Problem

I noticed a couple of oddities around the current `Metadata` class:
* It creates a `RdkafkaError` instance on a non-zero return from `rd_kafka_metadata` but doesn't raise it
* It returns Metadata for more than one topic even if a topic name is passed in to its initializer.

The only place we seem to use the Metadata class is in the [`Producer`](https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/producer.rb#L58), and it does an unnecessary `select` as it is getting more than one topic back from its request.

## The Solution

Fix the oddities, update the Producer so that it can assume it will only get a single topic's metadata back, and add explicit specs for the `Metadata` class.

### DIFF
--- a/lib/rdkafka/metadata.rb
+++ b/lib/rdkafka/metadata.rb
@@ -35,11 +35,11 @@ module Rdkafka
 
       @topics = Array.new(metadata[:topics_count]) do |i|
         topic = TopicMetadata.new(metadata[:topics_metadata] + (i * TopicMetadata.size))
-        Rdkafka::RdkafkaError.new(topic[:rd_kafka_resp_err]) unless topic[:rd_kafka_resp_err].zero?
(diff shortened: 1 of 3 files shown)


lib/rdkafka/metadata.rb
```diff
@@ -9,14 +9,15 @@ def initialize(native_client, topic_name = nil)
 
       ptr = FFI::MemoryPointer.new(:pointer)
 
-      # Retrieve metadata flag is 0/1 for single/multiple topics.
-      topic_flag = topic_name ? 1 : 0
+      # If topic_flag is 1, we request info about *all* topics in the cluster.  If topic_flag is 0,
+      # we only request info about locally known topics (or a single topic if one is passed in).
+      topic_flag = topic_name.nil? ? 1 : 0
```
Adithya-copart, line 14: Good catch!

PR #132, opened 2020-07-29
Enable zstd compression

jasonmartens:
We are using this plugin with Fluentd, and need to enable zstd compression for our log output. This PR works for us, but perhaps needs some additional changes to be compatible with everyone else? Let me know if this is something you would consider merging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ before_install:
       sudo apt-get install -y libpthread-stubs0-dev && \
       sudo apt-get install -y build-essential && \
       sudo apt-get install -y zlib1g-dev && \
+      sudo apt-get install -y libzstd-dev && \
       sudo apt-get install -y libssl-dev && \
       sudo apt-get install -y libsasl2-dev
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,19 +22,8 @@ rvm:
 before_install:
   - gem update --system
   - |
-    r_eng="$(ruby -e 'STDOUT.write RUBY_ENGINE')";
-    if [ "$r_eng" == "jruby" ]; then
-      sudo apt-get update && \
-      sudo apt-get install -y git && \
-      sudo apt-get install -y libpthread-stubs0-dev && \
-      sudo apt-get install -y build-essential && \
-      sudo apt-get install -y zlib1g-dev && \
-      sudo apt-get install -y libzstd-dev && \
-      sudo apt-get install -y libssl-dev && \
-      sudo apt-get install -y libsasl2-dev
-    else
-      apt-get update && apt-get install -y git build-essential libpthread-stubs0-dev zlib1g-dev libzstd-dev libssl-dev libsasl2-dev
-    fi
+    sudo apt-get update && \
+    sudo apt-get install -y git build-essential libpthread-stubs0-dev zlib1g-dev libzstd-dev libssl-dev libsasl2-dev
 
 before_script:
   - docker-compose up -d
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_install:
       sudo apt-get install -y libzstd-dev && \
       sudo apt-get install -y libssl-dev && \
       sudo apt-get install -y libsasl2-dev
-    elif
+    else
       apt-get update && apt-get install -y git build-essential libpthread-stubs0-dev zlib1g-dev libzstd-dev libssl-dev libsasl2-dev
     fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: ruby
-dist: xenial
+dist: bionic
 os: linux
 
 services:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
-
-sudo: false
+dist: xenial
+os: linux
 
 services:
   - docker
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -22,16 +22,8 @@ rvm:
...
-      sudo apt-get update && \
-      sudo apt-get install -y git && \
-      sudo apt-get install -y libpthread-stubs0-dev && \
-      sudo apt-get install -y build-essential && \
-      sudo apt-get install -y zlib1g-dev && \
-      sudo apt-get install -y libssl-dev && \
-      sudo apt-get install -y libsasl2-dev
-    fi
+    sudo apt-get update && \
+    sudo apt-get install -y git build-essential libpthread-stubs0-dev zlib1g-dev libzstd-dev libssl-dev libsasl2-dev
```
thijsc, line 26: Won't this break jRuby support?

jasonmartens: Previously this was enabled only for jRuby. The travis jRuby tests are still passing AFAICT. But I know very little about jRuby or how it might break, so it's worth trying it out.

jasonmartens:
@thijsc 
> Does this introduce new dependencies when using the gem? If so, are they generally available? Do they need to be documented?
Since this gem gets built using the process defined here, I don't *think* it will impact other dependencies? Clearly now zstdlib is required to build it, but it seems like that is taken care of by the build system here. But this is the first time I've seen this kind of native + ruby build, so I don't really know.

jasonmartens:
> I don't _think_ it will impact other dependencies? 

I will say that I did not have to update any other ruby dependencies in my projects aside from pointing them to our git source to install the gem.

thijsc:
> Since this gem gets built using the process defined here, I don't think it will impact other dependencies? Clearly now zstdlib is required to build it, but it seems like that is taken care of by the build system here. But this is the first time I've seen this kind of native + ruby build, so I don't really know.

I'm not sure if `zstdlib` is available on most/all Linux distributions by default, maybe you could look into finding the answer to that question?

jasonmartens:
> > Since this gem gets built using the process defined here, I don't think it will impact other dependencies? Clearly now zstdlib is required to build it, but it seems like that is taken care of by the build system here. But this is the first time I've seen this kind of native + ruby build, so I don't really know.
> 
> I'm not sure if `zstdlib` is available on most/all Linux distributions by default, maybe you could look into finding the answer to that question?

Here's a search on pkgs.org (I haven't used it before, seems legit?). https://pkgs.org/search/?q=zstd

Shows that most major distributions have it included (Ubuntu, Debian, RedHat, Alpine...).

thijsc:
Looks safe to add then, let's merge it and test it in the beta.

.travis.yml
thijsc: If the tests pass it should be ok, maybe these packages are available by default on bionic.

Adithya-copart: The gem is installed without the need for any dependencies in a MRI docker container with changes in this PR.

```
$ docker run -v $(pwd):/rdkafka -it ruby /bin/bash

root@8b7190045cd6:/# cd rdkafka/ && bundle && cd ext && rake
#=> Success

root@8b7190045cd6:~# uname -a
Linux 8b7190045cd6 4.19.76-linuxkit #1 SMP Tue May 26 11:42:35 UTC 2020 x86_64 GNU/Linux

```

Removing the conditional that installs the dependencies only in JRuby is not necessary but it does not harm either.

jasonmartens:
I don't have permissions to merge, and there are still changes requested.

grosser:
@thijsc can you please release this ?

thijsc:
I'm seeing some related build errors in #134:

```
  1) Rdkafka::Config configuration should allow configuring zstd compression
707     Failure/Error: raise ConfigError.new(error_buffer.read_string)
708     
709     Rdkafka::Config::ConfigError:
710       Invalid value "zstd" for configuration property "compression.codec"
711     # ./lib/rdkafka/config.rb:186:in `block (2 levels) in native_config'
712     # ./lib/rdkafka/config.rb:176:in `each'
713     # ./lib/rdkafka/config.rb:176:in `block in native_config'
714     # ./lib/rdkafka/config.rb:174:in `tap'
715     # ./lib/rdkafka/config.rb:174:in `native_config'
716     # ./lib/rdkafka/config.rb:138:in `producer'
717     # ./spec/rdkafka/config_spec.rb:95:in `block (3 levels) in <top (required)>'
```

https://travis-ci.org/github/appsignal/rdkafka-ruby/jobs/726285458

Need to look into this before we can release, there might be a dependency need.

grosser:
@jasonmartens know anything about that ?

grosser:
sounds like it might now require that these packages are installed ... so it should be optional, but it can't because the gem is compiled when installing ?

thijsc:
It looks like in this case it failed in runtime, so the compilation did succeed. It just can't start with this config enabled.

Could you find out on which distributions the required packages are installed by default?

jasonmartens:
It looks like the issues is that the branch https://github.com/geoff2k/rdkafka-ruby/tree/gt/update_to_librdkafka_1_5_0 hasn't merged the Travis config changes from this PR which installs the `zstd` packages.

thijsc:
I've made two test fixes in #139. Hopefully, that will make the build green so we can move on with the release.

jasonmartens:
@thijsc Ok, but the Travis changes should also be merged, so the test does what it's supposed to.

thijsc:
Isn't that included in 532afaa12aab6341bf4410324dee90055d9db1b6?

jasonmartens:
Yes, it is but I don't see the travis.yaml file changes in the branch with the failed build.

thijsc:
Think the branch needs a rebase, but I'l open a new pull since 1.5.2 came out in the meantime.

thijsc:
It looks like this issue is indeed fixed in `main`.

PR #133, opened 2020-07-30
actually raise an error when one occurs while reading metadata

breunigs:
It seems rather unintentional to create an error just to throw it away.

Additionally, a point release would be much appreciated, since I ran into the incorrect error usage (https://github.com/appsignal/rdkafka-ruby/commit/16a4497f59f4d369ebfb4ad0784161400fd96424)

### DIFF
--- a/lib/rdkafka/metadata.rb
+++ b/lib/rdkafka/metadata.rb
@@ -16,7 +16,7 @@ module Rdkafka
       result = Rdkafka::Bindings.rd_kafka_metadata(native_client, topic_flag, native_topic, ptr, 250)
 
       # Error Handling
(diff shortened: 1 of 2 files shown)


geoff2k:
I addressed the first of those in #128.   Good catch on the other two!

Mind if I add the other two (and the comment) to my PR with some test coverage?

geoff2k:
I've added the extra raises and adjusted the tests in https://github.com/appsignal/rdkafka-ruby/pull/128/commits/2a1ab905459b2cb8efe31b6ae5d13a5f4870ba9b.

breunigs:
Ah, your PR is much more comprehensive! Thus closing this one in favor of yours.

PR #135, opened 2020-09-12
Use API instead of shell commands to create test topics

geoff2k:
Now that appsignal#123 is merged and the API is in place do create and delete topics, use the API to create the topics needed for the integration tests.


thijsc (approved):
Nice work!

PR #137, opened 2020-10-15
Allow passing in topic configuration on create_topic

dezka:
With the changes introduced in PR: https://github.com/appsignal/rdkafka-ruby/pull/123, our use case is to be able to specify the topic configuration for topics upon creation. This helps us automate this out when spinning up new environments, etc, so we don't have to go back afterwards to set such config options.

I would like to propose that we add this ability to the `create_topic` method as shown.

Please let me know if you have any questions / concerns / suggestions for me to change, thanks!

### DIFF
--- a/lib/rdkafka/admin.rb
+++ b/lib/rdkafka/admin.rb
@@ -36,7 +36,7 @@ module Rdkafka
     # @raise [RdkafkaError] When the topic name is invalid or the topic already exists
     #
     # @return [CreateTopicHandle] Create topic handle that can be used to wait for the result of creating the topic
-    def create_topic(topic_name, partition_count, replication_factor)
+    def create_topic(topic_name, partition_count, replication_factor, topic_config={})
 
       # Create a rd_kafka_NewTopic_t representing the new topic
       error_buffer = FFI::MemoryPointer.from_string(" " * 256)
@@ -49,6 +49,16 @@ module Rdkafka
       )
       if new_topic_ptr.null?
         raise Rdkafka::Config::ConfigError.new(error_buffer.read_string)
+      end
+
+      unless topic_config.empty?
+        topic_config.each do |key, value|
+          Rdkafka::Bindings.rd_kafka_NewTopic_set_config(
+            new_topic_ptr,
+            key.to_s,
+            value.to_s
+          )
+        end
       end
 
       # Note that rd_kafka_CreateTopics can create more than one topic at a time
--- a/lib/rdkafka/admin.rb
+++ b/lib/rdkafka/admin.rb
@@ -52,7 +52,7 @@ module Rdkafka
         raise Rdkafka::Config::ConfigError.new(error_buffer.read_string)
       end
 
-      unless topic_config.empty?
+      unless topic_config.nil?
         topic_config.each do |key, value|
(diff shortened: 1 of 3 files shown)


lib/rdkafka/admin.rb
```diff
@@ -51,6 +52,16 @@ def create_topic(topic_name, partition_count, replication_factor)
         raise Rdkafka::Config::ConfigError.new(error_buffer.read_string)
       end
 
+      unless topic_config.empty?
+        topic_config.each do |key, value|
+          Rdkafka::Bindings.rd_kafka_NewTopic_set_config(
```
zainonrails, line 57: Wouldn't it make things consistent to have lower case methods names?

zainonrails (commented):
Added a suggestion to keep naming conventions consistent.

lib/rdkafka/admin.rb
dezka: My understanding is that it has to match exactly what the C function is called, so I'm not sure this can be lowered. Happy to be wrong though.

https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_admin.c#L1165

Other binding examples in the same situation: https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/bindings.rb#L253-L280

thijsc:

dezka:
@thijsc no worries, thanks for the heads up. I rebased and pushed. Please let me know if you need anything else.

lib/rdkafka/admin.rb
```diff
@@ -51,6 +52,16 @@ def create_topic(topic_name, partition_count, replication_factor)
         raise Rdkafka::Config::ConfigError.new(error_buffer.read_string)
       end
 
+      unless topic_config.empty?
```
geoff2k, line 55: Nitpick: won't the `.each` be a no-op if topic_config is `empty?`

spec/rdkafka/admin_spec.rb
```diff
@@ -14,6 +14,8 @@
   let(:topic_name)               { "test-topic-#{Random.new.rand(0..1_000_000)}" }
   let(:topic_partition_count)    { 3 }
   let(:topic_replication_factor) { 1 }
+  let(:topic_config)             { {"cleanup.policy" => "compact", "min.cleanable.dirty.ratio" => 0.8} }
+  let(:invalid_topic_config)     { {"cleeeeenup.policee" => "campact"} }
```
geoff2k, line 18: I think most spellcheck would have no suggestions for these misspellings!  😆

thijsc:
Could you rebase and force push this branch? That should trigger a build on Semaphore.

lib/rdkafka/admin.rb
dezka: Ah! yeah that makes sense. After looking at this a little more I'm going to replace this with a guard for nil.

spec/rdkafka/admin_spec.rb
dezka: Haha, yeah perhaps it's a little over the top 😄

ejoubaud:
Thanks a lot for this super useful change. It doesn't look like it's been released yet in a rubygems version. Any idea when it will be @thijsc ?

PR #138, opened 2020-10-19
Add note about resource consumption

ben:
This bit me not too long ago, and I thought I'd try to make it a bit more visible. With what I think is a very-common configuration, every producer created spawns a new thread that is never released, so `Config.new(...).producer.produce(...)` is a bad thing to be doing.


thijsc:
Thanks!

PR #141, opened 2020-12-07
Switch to Semaphore CI

thijsc:
Also fixes a few failing specs. Second attempt at switching from Travis.


thijsc:
I got it working this time. We're already paying for Semaphore at AppSignal, so it's nice to piggyback on that since Travis has a very limited free plan now.

PR #134, opened 2020-09-11
Update to librdkafka v1.5.0

geoff2k:
### The update

The most recent version of librdkafka v1.5.0 was released on [July 20th](https://github.com/edenhill/librdkafka/releases/tag/v1.5.0).  Both it and its two previous version ([v1.4.2](https://github.com/edenhill/librdkafka/releases/tag/v1.4.2) and [v1.4.4](https://github.com/edenhill/librdkafka/releases/tag/v1.4.4)) have multiple fixes we should incorporate.

To verify the SHA256 run the following:

```
$ curl -s https://codeload.github.com/edenhill/librdkafka/tar.gz/v1.5.0 | openssl dgst -sha256
f7fee59fdbf1286ec23ef0b35b2dfb41031c8727c90ced6435b8cf576f23a656
```

### The one regression

There was a single regression in a test in metadata_spec.rb that attempts to get metadata for a non-existent topic.

In 1.5.0, many updates were made related to the handling of unknown topics [note 1](https://github.com/edenhill/librdkafka/compare/v1.4.4...v1.5.0?w=1#diff-d66fa525591da8385b21fcab72e24f28R1461), [note 2 - "Upgrade Considerations"](https://github.com/edenhill/librdkafka/releases/tag/v1.5.0), so the exception raised in this case used to be `RD_KAFKA_RESP_ERR_LEADER_NOT_AVAILABLE`, but is now the more correct `RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART`.

### The bonus content

I added a few reminders to the `ext/README.md` to help future updaters.

PR #142, opened 2020-12-07
Update librdkafka to 1.5.2

thijsc:


geoff2k:
For this to succeed, we'll also need to move over the spec change I made in #134 that I mentioned in the summary:

https://github.com/appsignal/rdkafka-ruby/pull/134/files#diff-49c496906f9cb1caf05c348532cd184b1d401efeca804b87dc772995e1e62349

> **The one regression**
> There was a single regression in a test in metadata_spec.rb that attempts to get metadata for a non-existent topic.
> In 1.5.0, many updates were made related to the handling of unknown topics note 1, note 2 - "Upgrade Considerations", so the exception raised in this case used to be RD_KAFKA_RESP_ERR_LEADER_NOT_AVAILABLE, but is now the more correct RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART.

thijsc:
Closing in favour of  #146

PR #145, opened 2020-12-22
Allow any Callable Object for Callbacks

gremerritt:
When using callbacks (either for statistics or for producer delivery) it would be nice to be able to use something other than a Proc. For example, when handling statistics it may helpful to encapsulate handling logic into its own class.

This updates the type checking such that any object responding to `#call` can be used as a callback.

### DIFF
--- a/spec/rdkafka/producer_spec.rb
+++ b/spec/rdkafka/producer_spec.rb
@@ -41,31 +41,67 @@ describe Rdkafka::Producer do
       }.to raise_error(TypeError)
     end
 
-    it "should call the callback when a message is delivered" do
-      @callback_called = false
+    context "with a proc/lambda" do
+      it "should call the callback when a message is delivered" do
+        @callback_called = false
 
-      producer.delivery_callback = lambda do |report|
-        expect(report).not_to be_nil
-        expect(report.partition).to eq 1
-        expect(report.offset).to be >= 0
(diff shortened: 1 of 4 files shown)


spec/rdkafka/producer_spec.rb
```diff
@@ -21,38 +21,73 @@
       expect(producer.delivery_callback).to be_a Proc
```
geoff2k, line 21: Could we change this line to `expect(producer.delivery_callback).to respond_to :call` as we do in the `config_spec.rb` above, and add a test for a callable object?

Also, while you're here, would you mind changing the `puts stats` on line 18 to be `puts delivery_handle`?  Even though the code will never get executed, we probably should have something that would execute if it were asked to. :)

geoff2k (commented):
This looks like a great addition!

spec/rdkafka/producer_spec.rb
gremerritt: Heh, a little copy pasta 🍝. Updated the `puts` and added a spec for setting the callback.

thijsc:
Looks good! Could force push this? I've changed some settings, this should trigger a build on CI.

gremerritt:
@thijsc yep - done!

thijsc:
Thanks!

PR #151, opened 2021-02-26
Reduce memory allocations in Rdkafka::Producer#produce

jturkel:
Profiling parts of our event pipeline showed that ~5% of allocated memory came from some unnecessary array copying in `Rdkafka::Producer#produce`.


thijsc (approved):

thijsc:
Thanks!

PR #152, opened 2021-02-28
Bump librdkafka to 1.6.1

thijsc:


mensfeld:
@thijsc do  you want me to take it over?

thijsc:
That'd be great! Some test failures were introduced, but I'm not exactly sure why.

mrleedev:
I've tried this myself. [1.5.0 ]( https://github.com/edenhill/librdkafka/releases/tag/v1.5.0) introduces a change around auto topic creation, which is what causes some consumer specs to fail. 

Adding `KAFKA_AUTO_CREATE_TOPICS_ENABLE: "true"` to `docker-compose.yml` and adding `:"allow.auto.create.topics" => true,` into the config in `spec_helper` solved the problem, but not reliably. The test would pass if run in isolation, and only once, at which point it would fail again.

 [1.6.0](https://github.com/edenhill/librdkafka/releases/tag/v1.6.0) introduce some change around the broker admin API, specifially that `-1` means to use the default topic count on the broker: https://github.com/edenhill/librdkafka/blob/408f72593c43f727f4131310ba408378aacc1de6/src/rdkafka.h#L6140-L6141

I was able to fix that failing test by using a different invalid number.

Beyond that, I couldn't make any more progress because I get segfaults locally when `rd_destroy_kafka` is used. No point submitting the same PR twice, but maybe this info comes in handy :)

PR #155, opened 2021-04-02
Add bytes_threshold to Consumer#each_batch

gremerritt:
## Problem

On certain topics we have message that can vary greatly in size (anything up to 1MB), and in general the larger messages are more expensive to handle. In our current custom implementation of what is essentially `Consumer#each_batch` we handle this by setting a threshold on the total number of bytes consumed before returning a batch.

We would like to be able to adopt the provided `each_batch` method, however we need to ensure that we're not consuming too much data in a single batch.

## Solution

Introduce a `bytes_threshold` parameter to the `each_batch` method. As soon as we've consumed this much data, yield the batch.

cc @mgrosso

### DIFF
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -509,6 +509,8 @@ module Rdkafka
     #
     # @param max_items [Integer] Maximum size of the yielded array of messages
     #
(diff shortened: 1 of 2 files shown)


thijsc:
Thanks!

PR #158, opened 2021-04-28
GSSAPI Support

clutonik:
- Adding docker-compose-sasl.yml to support testing of SASL_SSL authentication. 
- Added --enable-ssl option for librdkafka installation.

Note:
- The docker-compose includes hostnames like kafka.confluent.io etc, please modify your /etc/hosts file to support these names. My /etc/hosts file:
- `127.0.0.1       localhost zookeeper.confluent.io kerberos.confluent.io consumer.confluent.io
    <my-ip> kafka.confluent.io`

- Use krb5.conf from scripts/secrets/krb5.conf to be able to test producer and consumer clients.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,7 @@ ext/librdkafka.*
 .yardoc
 doc
 coverage
+scripts/secrets/*
+!scripts/secrets/krb5.conf
+!scripts/secrets/zookeeper-jaas.conf
+!scripts/secrets/kafka-jaas.conf
--- a/README.md
+++ b/README.md
@@ -21,6 +21,12 @@ The most important pieces of a Kafka client are implemented. We're
 working towards feature completeness, you can track that here:
 https://github.com/appsignal/rdkafka-ruby/milestone/1
 
+# TODO
+
+- Setup SSL for kafka cluster running on docker using a script to create certs during build.
+- Consume and Produce messages using secure advertised listener. 
+- Enable SASL_SSL authentication.
+
 ## Installation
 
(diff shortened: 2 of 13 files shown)


thijsc:
Are there any changes needed in the library to support this scenario? Or are you just adding a test setup?

clutonik:
The only change needed from library is to install librdkafka using --enable-ssl flag but I add other scripts and docker-compose to be able to test it.

clutonik:
@thijsc Did you get a chance to review the changes?

thijsc:
I'm a bit on the fence about it. It's a lot of configuration in the repo that we're not going to be able to do maintenance on. Since there are no changes in the actual library this might be more of a concern for librdkafka itself.

Have you considered creating a separate repo with a full working example?

thijsc:
I'm closing this, at least for now. I think this would be great to put somewhere as an example repo.

clutonik:
@thijsc I have been a little busy at work lately and we actually shifted our implementation from SASL_SSL auth to mTLS but I will work on creating a repo to host example code required to test SASL_SSL auth with rdkafka-ruby. And I will open another PR to update README file in this repo. Tentative ETA would be end of next week.

skateman:
@thijsc would be okay to have a minimal version of this PR that just compiles the library with SSL support? We would like to use it under `racecar` where any other configuration is provided.

thijsc:
The library should already work with SSL, we use this in production actually. Is that not the case for you?

skateman:
Yeah, we're fine...sorry for the ping

thijsc:
No problem!

PR #161, opened 2021-07-16
Add configuration for error callbacks

leemeichin:
Adds configuration to use a global error callback as described in the C API here: https://docs.confluent.io/2.0.0/clients/librdkafka/rdkafka_8h.html#ace721ef3b7c22d0c111ec747ef039a90

The reason for this is that many errors are not surfaced in the consumer and are instead directed to the logger. Setting an error callback allows the consumer to react to these errors without inspecting the statistics payload.

### DIFF
--- a/lib/rdkafka/bindings.rb
+++ b/lib/rdkafka/bindings.rb
@@ -108,6 +108,8 @@ module Rdkafka
     attach_function :rd_kafka_conf_set_opaque, [:pointer, :pointer], :void
     callback :stats_cb, [:pointer, :string, :int, :pointer], :int
     attach_function :rd_kafka_conf_set_stats_cb, [:pointer, :stats_cb], :void
+    callback :error_cb, [:pointer, :int, :string, :pointer], :void
+    attach_function :rd_kafka_conf_set_error_cb, [:pointer, :error_cb], :void
 
     # Log queue
     attach_function :rd_kafka_set_log_queue, [:pointer, :pointer], :void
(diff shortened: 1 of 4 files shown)


lib/rdkafka/bindings.rb
```diff
@@ -146,6 +148,15 @@ class TopicPartitionList < FFI::Struct
       0
     end
 
+    ErrorCallback = FFI::Function.new(
+      :void, [:pointer, :int, :string, :pointer]
+    ) do |_client_prr, err_code, reason, _opaque|
```
geoff2k, line 153:
```suggestion
    ) do |_client_ptr, err_code, reason, _opaque|
```

spec/rdkafka/bindings_spec.rb
```diff
@@ -100,4 +100,28 @@
...
+    context "with an error callback" do
+      before do
+        Rdkafka::Config.error_callback = lambda do |error|
+          $received_error = error
+        end
+      end
+
+      it "should call the error callback with an Rdkafka::Error" do
+        Rdkafka::Bindings::ErrorCallback.call(nil, 8, "Broker not available", nil)
+        expect($received_error.code).to eq(:broker_not_available)
```
geoff2k, line 122: I scratched my head for awhile wondering where this symbol came from before remembering that we call `rd_kafka_err2name` to turn the 8 into `: broker_not_available`. 😆

geoff2k (commented):
Thank you for doing this!

spec/rdkafka/bindings_spec.rb
mrleedev: Yeah, I pulled an example out of this: https://github.com/edenhill/librdkafka/blob/58aa3ab832172770b1b690a06900df9c297b135c/src/rdkafka.h#L268-L618

lib/rdkafka/bindings.rb
mrleedev: Woops, good spot :)

mrleedev:
Not sure why the Java step is failing in CI when it previously worked. It seems unrelated but I can't restart the job either.

thijsc (approved):
Great work! Could you rebase this into a single commit? I'll hit merge straight away then :-).

mrleedev:
All done @thijsc :)

bougyman:
Any ETA on when this might get merged? We need it for our implementation.

thijsc:
Merged! I'll do a release shortly too.

PR #146, opened 2020-12-22
Update to librdkafka 1.5.3

geoff2k:
_This supercedes #142._

### The update

1.5.3, the most recent version of librdkafka v1.5.x was released on [December 9th](https://github.com/edenhill/librdkafka/releases/tag/v1.5.3).  It and its previous versions have multiple fixes we should incorporate.

To verify the SHA256 run the following:

```
$ curl -s https://codeload.github.com/edenhill/librdkafka/tar.gz/v1.5.3 | openssl dgst -sha256
2105ca01fef5beca10c9f010bc50342b15d5ce6b73b2489b012e6d09a008b7bf
```

### The one regression

There was a single regression in a test in metadata_spec.rb that attempts to get metadata for a non-existent topic.

In 1.5.0, many updates were made related to the handling of unknown topics [note 1](https://github.com/edenhill/librdkafka/compare/v1.4.4...v1.5.0?w=1#diff-d66fa525591da8385b21fcab72e24f28R1461), [note 2 - "Upgrade Considerations"](https://github.com/edenhill/librdkafka/releases/tag/v1.5.0), so the exception raised in this case used to be `RD_KAFKA_RESP_ERR_LEADER_NOT_AVAILABLE`, but is now the more correct `RD_KAFKA_RESP_ERR_UNKNOWN_TOPIC_OR_PART`.

### The bonus content

I added a few reminders to the `ext/README.md` to help future updaters.#146

PR #165, opened 2021-08-17
Upgrade to librdkafka 1.5.0

malandrina:
## Why?

My team maintains a stream processing application that uses [Racecar](https://github.com/zendesk/racecar), which depends on rdkafka-ruby. Recently we learned that a known issue with librdkafka 1.4.0 (https://github.com/edenhill/librdkafka/issues/2830) negatively impacts the performance of our application.

This PR proposes updating rdkafka-ruby to use librdkafka 1.5.0, which includes a fix for the issue referenced above.

## How?

We referenced prior art and discussion in https://github.com/appsignal/rdkafka-ruby/pull/152 and made the following changes:

- Set `Rdkafka::LIBRDKAFKA_VERSION` to 1.5.0
- Set `Rdkafka::LIBRKAFKA_SOURCE_SHA246` to the SHA associated with librdkafka v.1.5.0, which we found in the librdkafka release notes: https://github.com/edenhill/librdkafka/releases?after=v1.6.0-PRE4
- Update specs to explicitly create topics in the setup phase instead of relying on Kafka to automatically create the topics
- Update a test assertion to match against the new expected error message

### Testing

We followed the instructions for setting up the rdkafka-ruby development environment in the README and confirmed tests pass locally.

### Considerations

As a stopgap, my team is currently using gems built from our own forks of the racecar and rdkafka-ruby repos in production. I’m hoping that the changes in this PR will be merged and released, and then my next step will be to open a PR on the zendesk/racecar repo proposing an upgrade to the rdkafka-ruby dependency.

Thanks in advance for your time! 🙇‍♀️


geoff2k:
These changes looks oddly familiar... 😉 (#134).

That said, can you say more about the change where we don't rely on Kafka to auto-create the topics for us?   My understanding was that the [confluent docker images](https://github.com/appsignal/rdkafka-ruby/blob/main/docker-compose.yml#L13) we are using to test against are configured for auto-topic creation.  Has that changed?

thijsc:
I haven't found the time lately to look into the failing specs for #152. Any reason you went specifically for 1.5.0?

malandrina:
@geoff2k thanks for linking to #134. I also quickly reviewed #142 and #146, which I hadn’t seen before. I see that they were all closed without merging in favor of #152.

> can you say more about the change where we don't rely on Kafka to auto-create the topics for us? My understanding was that the confluent docker images we are using to test against are configured for auto-topic creation. Has that changed?

Good question! I haven’t checked the configuration of the confluent Docker images. Our decision to update the tests to avoid relying on auto-creation of topics came out of a desire to avoid investigating and debugging the flakiness `@leemeichin` described in https://github.com/appsignal/rdkafka-ruby/pull/152#issuecomment-882055825.

With respect to @thijsc’s question

> Any reason you went specifically for 1.5.0?

Yes — librdkafka 1.5.0 includes a fix for https://github.com/edenhill/librdkafka/issues/2830, an issue with 1.4.0 that is is impacting the performance of our application. At present my team has no need to run a more recent version of librdkafka.

malandrina:
Thanks for merging this @thijsc!

PR #167, opened 2021-08-26
Use Homebrew prefix when setting OpenSSL compile flags

robbiepaul:
Homebrew on Apple Silicon/ARM uses `/opt/homebrew` [path prefix ](https://docs.brew.sh/FAQ#why-should-i-install-homebrew-in-the-default-location)instead of `/usr/local` (used by Intel Macs).

The [issue has been identified in the `zendesk/racecar` repo](https://github.com/zendesk/racecar/issues/255) but I believe this commit should fix the problem.

[1]: https://docs.brew.sh/FAQ#why-should-i-install-homebrew-in-the-default-location
[2]: https://github.com/zendesk/racecar/issues/255

### DIFF
--- a/ext/Rakefile
+++ b/ext/Rakefile
@@ -25,9 +25,9 @@ task :default => :clean do
 
   # Use default homebrew openssl if we're on mac and the directory exists
   # and each of flags is not empty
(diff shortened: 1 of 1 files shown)


thijsc:
Thanks!

Issue #82, opened 2019-08-09
Non global logger and statistics callback


mensfeld:
Majority of the settings of this lib are per instance based.

Is there any reason why `logger` and `statistics_callback` aren't? It is really problematic for systems where you may have several instances of rdkafka running for producing and consuming with different policies around instrumentation.

thijsc:
Looking at the docs it does seem like both are scoped on the client. So it should be possible to move them to a config's instance. I have some other issues with higher priority, will take a look at this after those. Also open to a pull of course.

thijsc:
Closing this one since we don't currently need this. If anybody needs this do reopen.

mensfeld:
> If anybody needs this do reopen.

I do not have a way to reopen :(

thijsc:
Are you interested in working on this?

mensfeld:
@thijsc yes. Just leave it open for 2-3 more weeks. I'm finishing Karafka 2.0 and then I will have some time to adjust things on rdkafka side :)

mensfeld:
Closing as I was able to bypass this limitation without changing the design. If anyone needs this, just ping me.

trickleup:
@mensfeld - running into the same issue that you describe, of needing to have multiple instances of producers/consumers, and to be able to isolate the statistics_callbacks between those instances.

> I was able to bypass this limitation without changing the design. If anyone needs this, just ping me.

Do you mind sharing the gist of what you did to achieve this?

mensfeld:
@trickleup I'm on a holiday and I plan to backport if from karafka and waterdrop to rdkafka once back (2 weeks). 

If you are looking for multiple instances I would recommend going with the wrapper I did waterdrop: https://github.com/karafka/waterdrop as it tackles those cases. The code for this is also somewhere there. Writing from mobile so can't locate the exact files sorry :pray:

jychen7:
> I plan to backport if from karafka and waterdrop to rdkafka once back (2 weeks).

👍

mensfeld:
@jychen7 ok clearly it is taking me more than that but my original solution still stands and I still have plans to merge this and other things to the upstream

PR #170, opened 2021-09-20
Use non-global statistics and error callbacks

mensfeld:
Hey, giving a shot to fix this: https://github.com/appsignal/rdkafka-ruby/issues/82

This PR introduces a per config interface to provide callback objects for statistics and errors. It allows for a better separation of responsibilities when several producers or consumers are used.

It also includes specs to make sure, that when we have more than one producer / consumer, their callbacks do not interact with each other.


mensfeld:
Failing sub-specs seem unrelated.

mensfeld:
@thijsc ping pretty please :pray:

thijsc:
I think this should work well. Did you test what happens if you mix create consumers/producers from both single and multiple configs?

mensfeld:
> I think this should work well. Did you test what happens if you mix create consumers/producers from both single and multiple configs?

I did. I just wanted to get a confirmation, that the direction I'm heading is ok (closure around FFI). I will add appropriate specs to handle the cases you've asked.

mensfeld:
Please don't merge yet despite :green_heart: I'm going to rework the log callback and add specs to make sure that callbacks do not interact with each other.

mensfeld:
Closing this one unless anyone else needs it (then I can continue working on it). I was able to bypass this by getting the instance name from librdkafka and using it to filter out the callbacks that are published, hence this change won't be needed by me anymore.

Ruto8:
@mensfeld, thoughts about re-opening this? We want to use the stats callback, but the global nature of it stops us. I saw the workaround in Waterdrop, but I'm reluctant to implement it myself.

I was thinking about contributing, but your PR is everything we need.

mensfeld:
@Ruto8 not at the moment. There are couple of things related to global state and defaults that I want to tackle altogether in the near future. Just need to finalize some other work. I would suggest for now to go with waterdrop solution and we can coop on a proper PR in the near future. waterdrop solution is 100% reliable and stable. Any change to callbacks logic here needs to be extremely well tested (which I do) and I must be super cautious about it.

Issue #162, opened 2021-07-21
Questions about Producer termination sequence


caspg:
According to [this doc](https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#producer) the proper termination sequence for Producer is:

```c
     /* 1) Make sure all outstanding requests are transmitted and handled. */
     rd_kafka_flush(rk, 60*1000); /* One minute timeout */

     /* 2) Destroy the topic and handle objects */
     rd_kafka_topic_destroy(rkt);  /* Repeat for all topic objects held */
     rd_kafka_destroy(rk);
```

In `rdkafka-ruby` we are just calling `rd_kafka_destroy` [click](https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/producer.rb#L50). Are we not calling `rd_kafka_flush` because of [this condition in polling loop](https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/producer.rb#L23)? Can we assume that all messages were delivered when calling `producer#close`?

What about `rd_kafka_topic_destroy`?

Thanks

PR #172, opened 2021-10-27
Extract lower level kafka client from producer to decouple finalizer logic and close logic

jvortmann:
## Summary

This fixes the following error
```ruby
FrozenError:
       can't modify frozen String
```
that we start seeing on our apps starting on version `v0.10.0` due the use of a `String` on the `ObjectSpace.define_finalizer` on https://github.com/appsignal/rdkafka-ruby/blob/319b929dc1d524c8a31542aaa0942342415f7ff8/lib/rdkafka/producer.rb#L19 from https://github.com/appsignal/rdkafka-ruby/commit/b7c790b21427ad419d8144d6c56b0655e2148adc. This happens when instantiating a new producer, probably due to the presence of:
```ruby
# frozen_string_literal: true
```

It does not change the previous behavior of not closing when the producer instance was de-referenced as discussed in https://github.com/appsignal/rdkafka-ruby/pull/115 but still removes the warning that was the origin of https://github.com/appsignal/rdkafka-ruby/commit/b7c790b21427ad419d8144d6c56b0655e2148adc.

The issue mentioned on https://www.mikeperham.com/2010/02/24/the-trouble-with-ruby-finalizers/ and by the `ObjectSpace.define_finalizer` documentation (https://ruby-doc.org/core-3.0.2/ObjectSpace.html#method-c-define_finalizer) that was the origin of the `warning` that meant to be removed was in fact related to the second parameter, not the first. So using a `String` as the first parameter removed the warning but was used to indirectly close the native client (after the `String` was GCed, not the Producer instance per-se). By the documentation mentioned, the first parameter is the object being finalized and the finalizer is meant to be ran after it is GCed.

This solution follows the second suggestion on the previous mentioned post and documentation by wrapping the native-related code (native client and thread) into a Class, moving the close method there and using that as a callable finalizer by defining and alias to the `close` method as `call` (expected by `ObjectSpace.define_finalizer`'s second parameter).
It also exposes the native client and `closed?` methods to avoid direct access and do not change behavior.

Additionally, in moving the polling thread into the wrapper class, the `closing` variable was turn into a thread local variable, as it had no access outside of the class or `close` method.

Ideally, in the future this inner class can be move up (out of the producer) and injected directly. This will allow the `ObjectSpace.define_finalizer` to receive the injected instance and avoid the self reference that is the impediment to GC when the producer instance is de-referenced. For that additional method that reference the native client are needed to be implemented there (like `partition_count`, the call to `Rdkafka::Bindings.rd_kafka_producev` and even the closed check. As these would make this PR bigger, they were not done yet.

-> Did this on the latest commit, but the GC does not remove the producer reference because the opaque is holding a reference to it on: https://github.com/jvortmann/rdkafka-ruby/blob/b51dc0fdaac1fe18cc1280d6accbd1532790dff2/lib/rdkafka/config.rb#L183
If this line is removed, doing the following will close and GC the producer (via console):
```
config = {:"api.version.request"=>false, :"broker.version.fallback"=>"1.0", :"bootstrap.servers"=>"localhost:9092", :"group.id"=>"ruby-test-287694", :"auto.offset.reset"=>"earliest", :"enable.partition.eof"=
>false}
p1 = Rdkafka::Config.new(config).producer
ObjectSpace.each_object(Rdkafka::Producer).count # => 1
p1 = nil
ObjectSpace.garbage_collect
ObjectSpace.each_object(Rdkafka::Producer).count # => 0
```

---
### Additional Suggestions
It would really good to have the versions tagged as git tags to easily navigate a particular version on Github.
I've created the following on my fork:
- `0.8.0`: `git tag v0.8.0 4185876`
- `0.8.1`: `git tag v0.8.1 90efb49`
- `0.9.0`: `git tag v0.9.0 380678f`
- `0.10.0`: `git tag v0.10.0 90285c9`
- `0.11.0rc1`: `git tag v0.11.0rc1 5c857d0` -> so we can cut a `release-candidate` version to use internally

---
### Unrelated additions
A couple of more changes here done prior to working on the solution:
- The versions of `zookeeper` and `kafka` images on the `docker-compose.yml` were fixes because the `latest` wouldn't boot, just `exit 1`
- `Guard` was added to run the spec on save (`bundle exec guard`)
- `bin/console` was added to open an `irb` session with code loaded and spike things if needed
- `focus` rspec options was added so individual spec could be ran with `fit`, `fcontext` or `:focus` params.

### DIFF
--- a/spec/rdkafka/producer/client_spec.rb
+++ b/spec/rdkafka/producer/client_spec.rb
@@ -1,0 +1,144 @@
+require "spec_helper"
+
+describe Rdkafka::Producer::Client do
+  let(:native) { double }
+  let(:closing) { false }
+  let(:thread) { double(Thread) }
+
+  subject(:client) { described_class.new(native) }
+
+  before do
+    allow(Rdkafka::Bindings).to receive(:rd_kafka_poll).with(native, 250)
+    allow(Rdkafka::Bindings).to receive(:rd_kafka_outq_len).with(native).and_return(0)
+    allow(Rdkafka::Bindings).to receive(:rd_kafka_destroy)
+    allow(Thread).to receive(:new).and_return(thread)
+
+    allow(thread).to receive(:[]=).with(:closing, anything)
+    allow(thread).to receive(:join)
+    allow(thread).to receive(:abort_on_exception=).with(anything)
+  end
+
+  context "defaults" do
+    it "sets the thread to abort on exception" do
+      expect(thread).to receive(:abort_on_exception=).with(true)
+
+      client
+    end
+
+    it "sets the thread `closing` flag to false" do
+      expect(thread).to receive(:[]=).with(:closing, false)
+
+      client
+    end
+  end
+
+  context "the polling thread" do
+    it "is created" do
+      expect(Thread).to receive(:new)
+
+      client
+    end
+
+    it "polls the native with default 250ms timeout" do
+      polling_loop_expects do
+        expect(Rdkafka::Bindings).to receive(:rd_kafka_poll).with(native, 250)
+      end
+    end
+
+    it "check the out queue of native client" do
+      polling_loop_expects do
+        expect(Rdkafka::Bindings).to receive(:rd_kafka_outq_len).with(native)
+      end
+    end
+  end
+
+  def polling_loop_expects(&block)
+    Thread.current[:closing] = true # this forces the loop break with line #12
+
+    allow(Thread).to receive(:new).and_yield do |_|
+      block.call
+    end.and_return(thread)
+
+    client
+  end
+
+  it "exposes `native` client" do
+    expect(client.native).to eq(native)
+  end
+
+  context "when client was not yet closed (`nil`)" do
+    it "is not closed" do
+      expect(client.closed?).to eq(false)
+    end
+
+    context "and attempt to close" do
+      it "calls the `destroy` binding" do
+        expect(Rdkafka::Bindings).to receive(:rd_kafka_destroy).with(native)
+
+        client.close
+      end
+
+      it "indicates to the polling thread that it is closing" do
+        expect(thread).to receive(:[]=).with(:closing, true)
+
+        client.close
+      end
+
+      it "joins the polling thread" do
+        expect(thread).to receive(:join)
+
+        client.close
+      end
+
+      it "closes and unassign the native client" do
+        client.close
+
+        expect(client.native).to eq(nil)
+        expect(client.closed?).to eq(true)
+      end
+    end
+  end
+
+  context "when client was already closed" do
+    before { client.close }
+
+    it "is closed" do
+      expect(client.closed?).to eq(true)
+    end
+
+    context "and attempt to close again" do
+      it "does not call the `destroy` binding" do
+        expect(Rdkafka::Bindings).not_to receive(:rd_kafka_destroy)
+
+        client.close
+      end
+
+      it "does not indicate to the polling thread that it is closing" do
+        expect(thread).not_to receive(:[]=).with(:closing, true)
+
+        client.close
+      end
+
+      it "does not join the polling thread" do
+        expect(thread).not_to receive(:join)
+
+        client.close
+      end
+
+      it "does not close and unassign the native client again" do
+        client.close
+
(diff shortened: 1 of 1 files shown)


jvortmann:
Not using `self` on the `ObjectSpace.define_finalizer` tries to remove the
warning but let the producer finalization happen indirectly: when, and
only when, the `string` is GCed is the finalizer called. That seems more
important than the issue we faced.

The documentation[^finalizer-doc] of the `ObjectSpace.define_finalizer`
indicates that the first parameter is the object that need a finalizer
after being destroyed. Hence, using something other that `self` is
actually running the finalizer based on the life-cycle of another object,
the `@id` in this case.

[^finalizer-doc]: https://ruby-doc.org/core-3.0.2/ObjectSpace.html#method-c-define_finalizer

To exemplify this case you can run the following:
```
# take the above `config`
producer = config.producer
producer.instance_variable_get(@:id) # some uuid
producer.instance_variable_set(@:id, nil) # remove the reference
ObjectSpace.garbage_collect # -> this calls the finalizer even though producer still being used
```
This example demonstrates that the producer's finalizer depends on the life-cycle
of another object instead of itself.

Additionally, extracting the inner lower level details of thread polling
and closing to a separate class cleans up the producer considerably, as
considered on #160 and could help to easily introduce the suggestions of
#162 (rd_kafka_flush) there.
This also moves all code that need finalizing to a separate class and
injects it as a dependency, decoupling the lower level `native kafka`
related code from the more higher level `producer` code.

The potential memory leak happens in either solution when using
`config.producer` as this factory method hold a reference to the created
producer on the `Opaque` that is kept on the `Config.opaques` class
variable and prevent the GC from freeing the producer.
Calling `config.producer` seems seems to be the expected way of creating
producers instead of `Producer.new`. The referenced solution by Mike[^mike-blog]
also holds a reference due to the `proc` inline.

[^mike-blog]: https://www.mikeperham.com/2010/02/24/the-trouble-with-ruby-finalizers/

---

About the error specifically: that happened on our CI more than a month
ago, and we applied the fix explained here when detected. Unfortunately,
we can not reproduce the entire stacktrace right now, only the aforementioned
`Frozen String` error.

The code is basically this:
```ruby
# frozen_string_literal: true

class Wrapper
  def initialize(config)
    @client = config.producer
  end
end

# remaining settings ommited
config = Rdkafka::Config.new({ :"bootstrap.servers" => "array-of-host,separated", :"client.id" => "client_name"})

Wrapper.new(config) # this raises the error pointing to the line ObjectSpace.define_finalizer(@id, proc { close })
```
There is an hypothesis that the test setup has some influence but making
the finalizer rely on the producer directly rather than indirectly
through another object is a another main benefit and prevents future
hard-to-debug issues.

mensfeld:
> Not using self on the ObjectSpace.define_finalizer tries to remove the
warning but let the producer finalization happen indirectly

I am aware of it. The intent was to trigger the finalization upon exit. Now that I think of if an `at_exit` could potentially work as well (though there may be some trap context issues).

> Hence, using something other that self is actually running the finalizer based on the life-cycle of another object, the @id in this case.

I am aware of this. That's why the id is owned by the producer and limited to internal usage only. 

Don't get me wrong: Your solution while more complex works :) so I don't have a problem with it atl! I just want to understand who/what/when is mutating something internally as this may cause other hard to debug issues in the future that no-one anticipated.

thijsc:
Could you rebase this? And maybe split out the rspec setup changes to a separate pull?

jvortmann:
Hey @thijsc I've extracted the other changes to other PRs:
- [ ] #177 
- [ ] #178 
- [ ] #179

jvortmann:
@thijsc I've also rebased and moved the class out to a separate file to make the change cleaner.

jvortmann:
Hey @thijsc, I've have rebased on top of latest `main` and added a spec for the extracted client.

Does it make sense now to rename this PR to a extraction instead?

CarlosGabaldon:
👋 @thijsc any updates on this?

thijsc:
I've released this as `0.12.0.beta.0`, could you try that version out?

jvortmann:
@thijsc We were using a forked version with that internally, gonna try this beta one out.

Issue #171, opened 2021-10-19
Ignore `partitioner` config when I pass `partition_key` to `Producer#produce`


joker1007:
When `Producer#produce` method receives `partition_key`, it uses `Rdkafka::Bindings.partitioner`.
The partitioner algorithm of `Rdkafka::Bindings.partitioner` is `consitent_random` and is not configurable.

Because of it, when I used rdkafka with `{partitioner: "murmur2_random"}` config, the partition ID is different for the same key depending on whether Producer#produce receives the partition_key or not.
I think that the behavior is confusing.

I suggest `Rdkafka::Bindings.partitioner` takes `partitioner` config into account.
Or a new configuration parameter may be suitable.

PR #173, opened 2021-11-01
Use specified partitioner for partition key

abicky:
Rdkafka allows us to specify the partition key but always calculates the partition using consistent_random partitioner even if a partitioner is specified in the configuration.
This PR makes it use the specified partitioner and also resolves https://github.com/appsignal/rdkafka-ruby/issues/171.

### DIFF
--- a/lib/rdkafka/bindings.rb
+++ b/lib/rdkafka/bindings.rb
@@ -246,14 +246,21 @@ module Rdkafka
     attach_function :rd_kafka_conf_set_dr_msg_cb, [:pointer, :delivery_cb], :void
 
     # Partitioner
-    attach_function :rd_kafka_msg_partitioner_consistent_random, [:pointer, :pointer, :size_t, :int32, :pointer, :pointer], :int32
+    PARTITIONERS = %w(random consistent consistent_random murmur2 murmur2_random fnv1a fnv1a_random).each_with_object({}) do |name, hsh|
+      method_name = "rd_kafka_msg_partitioner_#{name}".to_sym
+      attach_function method_name, [:pointer, :pointer, :size_t, :int32, :pointer, :pointer], :int32
+      hsh[name] = method_name
+    end
 
-    def self.partitioner(str, partition_count)
+    def self.partitioner(str, partition_count, partitioner_name = "consistent_random")
       # Return RD_KAFKA_PARTITION_UA(unassigned partition) when partition count is nil/zero.
       return -1 unless partition_count&.nonzero?
 
       str_ptr = FFI::MemoryPointer.from_string(str)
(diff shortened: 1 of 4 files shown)


lib/rdkafka/bindings.rb
```diff
@@ -246,14 +246,21 @@ class TopicPartitionList < FFI::Struct
...
     # Partitioner
-    attach_function :rd_kafka_msg_partitioner_consistent_random, [:pointer, :pointer, :size_t, :int32, :pointer, :pointer], :int32
+    PARTITIONERS = %w(random consistent consistent_random murmur2 murmur2_random fnv1a fnv1a_random).each_with_object({}) do |name, hsh|
+      method_name = "rd_kafka_msg_partitioner_#{name}".to_sym
+      attach_function method_name, [:pointer, :pointer, :size_t, :int32, :pointer, :pointer], :int32
+      hsh[name] = method_name
+    end
 
-    def self.partitioner(str, partition_count)
+    def self.partitioner(str, partition_count, partitioner_name = "consistent_random")
```
thijsc, line 255: Maybe maken this `nil` by default and skip setting it? Then the default behaviour stays the same.

spec/rdkafka/producer_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "spec_helper"
+require "zlib"
```
thijsc, line 2: This can be removed after a rebase.

thijsc (changes requested):
Can you rebase?

lib/rdkafka/bindings.rb
abicky: I'm sorry but I'm afraid I don't understand your intention. I set the default value "consistent_random" for the users who call this method directly like `Rdkafka::Bindings.partitioner(str)`, and I think the default behavior, including such a case, hasn't changed.
Which do you suggest?

```ruby
def self.partitioner(str, partition_count, partitioner_name = nil)
  ...    
end
```

or

```ruby
def self.partitioner(str, partition_count, partitioner_name)
  ...    
end
```

spec/rdkafka/producer_spec.rb
abicky: I've rebased the branch and removed it.

abicky:
> Can you rebase?

I'm sorry the branch was too old. I've rebased it.

abicky:
@thijsc Can you review the changes?

thijsc:
Could you rebase once more? Sorry for the delay reviewing this.

abicky:
@thijsc I've rebased the branch onto the latest "main."

thijsc:
Thanks!

PR #187, opened 2022-03-15
Fix red build

leonmaia:
Description is WIP.

Fixes red build as discussed in extent on https://github.com/appsignal/rdkafka-ruby/pull/185.


leonmaia:
@thijsc It seems this solves the issues with the build, can you have a look?

jvortmann:
@leonmaia this is essentially what is on #189 but with a wrong setup (it is using a `consumer` config instead of producer.

It think we can keep only the later one for @thijsc to see and close this one.
What you think?

PR #191, opened 2022-03-30
Use RUBYARCHDIR for library location if present

### LINKED ISSUE #182: `librdkafka.so` is not cached by Bundler
Copying `librdkafka.so` into `ext` breaks the Bundler cache; it's expected that native extensions compile artifacts to the gem's [`extension_dir`](https://github.com/rubygems/rubygems/blob/20d4957649fb7eaf4d57c220f5302ce2f4ee05ab/lib/rubygems/basic_specification.rb#L94-L108), not into `ext` directly; the result is that Bundler caches the (empty) extension directory, and does not include `ext/librdkafka.so` in the cache.

This means that reinstalling the gem fails if it is present in the Bundler cache.

To reproduce:

```ruby
# Gemfile

source "https://rubygems.org"
gem "rdkafka"
```

```console
$ bundle config set path vendor/bundle
$ bundle install
$ rm -rf vendor/bundle
$ bundle install
```

After the first install, `vendor/bundle/ruby/2.7.0/gems/rdkafka-0.11.1/ext/librdkafka.so` exists. After the second, it does not, and attempting to use the gem fails.

Related issue: https://github.com/mimemagicrb/mimemagic/issues/121
And related fix: https://github.com/mimemagicrb/mimemagic/pull/136

As you can see in the related fix it is quite easy to resolve, you just have to write the compiled artifact to the directory specified by the `RUBYARCHDIR` environment variable rather than to `ext` itself.

thijsc:
Fix #182

### DIFF
--- a/ext/Rakefile
+++ b/ext/Rakefile
@@ -28,16 +28,25 @@ task :default => :clean do
     from_extension = 'so.1'
     to_extension = 'so'
   end
+
+  # Get path to the compiled library
   lib_path = File.join(File.dirname(__FILE__), "ports/#{recipe.host}/librdkafka/#{Rdkafka::LIBRDKAFKA_VERSION}/lib/librdkafka.#{from_extension}")
-  FileUtils.mv(lib_path, File.join(File.dirname(__FILE__), "librdkafka.#{to_extension}"))
+
+  # Get target dir
+  target_dir = ENV["RUBYARCHDIR"] || File.dirname(__FILE__)
+
+  # Move the compliled library there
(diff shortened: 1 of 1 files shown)


leonmaia:
With this change we are now breaking this line https://github.com/appsignal/rdkafka-ruby/blob/d973708d89dab78330f2609793262113e1d1430a/lib/rdkafka/bindings.rb#L18 @thijsc .

leonmaia:
@thijsc I'm reverting this for now in order to fix the latest release.

PR #194, opened 2022-04-12
Use RUBYARCHDIR for library location if present

leonmaia:
WIP 

previous revert: https://github.com/appsignal/rdkafka-ruby/pull/193

### DIFF
--- a/ext/Rakefile
+++ b/ext/Rakefile
@@ -28,16 +28,25 @@ task :default => :clean do
     from_extension = 'so.1'
     to_extension = 'so'
   end
+
+  # Get path to the compiled library
   lib_path = File.join(File.dirname(__FILE__), "ports/#{recipe.host}/librdkafka/#{Rdkafka::LIBRDKAFKA_VERSION}/lib/librdkafka.#{from_extension}")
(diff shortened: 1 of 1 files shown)


leonmaia:
I may be missing an easy solution here, but my main thought right now is that rubygems is pretty good in handling C extensions and this change would be quite straightforward if we were using it but when it comes to using FFI is on us to figure out. 

I failed to find a way to provide the correct path in the bindings app which would look something like `/Users/USER_NAME/.gem/extensions/OS_ARCHITECTURE/RUBY_VERSION/GEM_NAME_AND_VERSION`, example: `/Users/hello_world/.gem/extensions/x86_64-darwin-20/2.6.0/rdkafka-0.12.0.beta.3`

leonmaia:
I may be missing an easy solution here, but I can't really think of a way right now to provide the correct path in the bindings file, which would look something like `/Users/USER_NAME/.gem/extensions/ARCHITECTURE/RUBY_VERSION/GEM_NAME_AND_VERSION`, example: `/Users/hello_world/.gem/extensions/x86_64-darwin-20/2.6.0/rdkafka-0.12.0.beta.3/librdkafka.dylib`

thijsc:
Closing this for now, think we should re-think the approach.

Issue #195, opened 2022-04-22
No delivery handle reference in delivery callback


laurynas:
Delivery report is passed in the delivery callback, but it is impossible (?) to find a delivery handle by it. 

`produce` returns delivery handle and if we would like to wait for the callback to check the status, not it is not possible. The only option is `wait`. 

Looks like originally it was intended to pass delivery handle to the callback, because callback argument is named `delivery_handle`:
https://github.com/appsignal/rdkafka-ruby/blob/v0.12.0.beta.3/lib/rdkafka/producer.rb#L154

 I'm not sure what is the use of delivery report in the callback, if it can't be traced to the original producer call..

PR #196, opened 2022-04-22
Pass the delivery handle to the callback

gvisokinskas:
Added a small improvement to the producer that allows configuring callbacks that accept delivery handles as a second parameter to the callback. 

This is intended to address the issue described in https://github.com/appsignal/rdkafka-ruby/issues/195. Had to address some issues with callback arity to make this fully backwards compatible.

### DIFF
--- a/spec/rdkafka/producer_spec.rb
+++ b/spec/rdkafka/producer_spec.rb
@@ -49,6 +49,27 @@ describe Rdkafka::Producer do
         # Callback should have been called
         expect(@callback_called).to be true
       end
+
+      it "should provide handle" do
+        @callback_handle = nil
+
+        producer.delivery_callback = lambda { |_, handle| @callback_handle = handle }
+
+        # Produce a message
+        handle = producer.produce(
+          topic:   "produce_test_topic",
+          payload: "payload",
+          key:     "key"
+        )
+
+        # Wait for it to be delivered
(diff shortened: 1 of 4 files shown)


spec/rdkafka/producer_spec.rb
```diff
@@ -93,6 +114,36 @@ def call(report)
         expect(called_report.first.partition).to eq 1
         expect(called_report.first.offset).to be >= 0
       end
+
+      it "should provide handle" do
+        callback_handle = []
```
laurynas, line 119: Should be plural - `callback_handles`

gvisokinskas: I've followed the pattern from the previous test: `called_report = []`, the array is used as a "reference wrapper" of sorts to capture the provided handle.

laurynas: Ah, I see. Maybe it is ok for consistency. In general it it would make sense to name it plural, because it can be mistaken for callback handle itself.

leonmaia: Hi @gvisokinskas thanks for the PR! I agree with @laurynas, let's make it plural. We can change `called_report` to also use plural in another PR.

leonmaia (approved):
Thanks for contributing @gvisokinskas!

Issue #190, opened 2022-03-29
Segfaults thrown from producer running in Rails on Puma.


dezka:
**rdkafka-ruby version: 0.11.1
ruby 2.7.0, Rails 6, Puma 5.5.2**

We are running into a segmentation fault problem that is consistently thrown from [rd_kafka_msg_partitioner_consistent_random](https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/bindings.rb#L256) when calling `#produce` with a `partition_key` set. However the segfault situation itself is sporadic and random.

In this situation, we have an rdkafka producer running in Rails on Puma. Our running theory is that the rdkafka thread might be impacted by forking that Puma is doing. In our experience rdkafka-ruby runs into a segfault problem with forking and we’ve made it so our producers are lazy loaded (i.e. producer is starting well after the app is booted.) To try and squash this particular problem, we are running Puma config of `workers 4` and `threads 1, 1 (min, max)` to force Puma to be single threaded. This hasn’t changed a thing. But we are focusing now on how it’s the same error at the same C function every time with the same segfault message like: `Segmentation fault at 0x0000000000000000`. This hexadecimal value represents nil/null if I recall correctly. So something such as a pointer used in this C func is nil’d out by the time this runs.

Side note: [it seems we are not passing in the topic](https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/bindings.rb#L256) for the function to do the lookup upon, which is the first arg [here](https://github.com/edenhill/librdkafka/blob/0261c86228e910cc84c4c7ab74e563c121f50696/src/rdkafka.h#L2731). I don’t think this is necessarily causing the problem, but other than some other thread being killed, this is the only thing I can imagine being `nil` at the time of execution that it is complaining about. I’m sure I’m missing something here though.

Full stack trace: https://gist.github.com/dezka/92c6747bf1ca2b0a01f407a7c52f442f

mensfeld:
This is an issue for a long time: I bypassed that in WaterDrop by detecting a fork and raising an exception because of that:

https://github.com/karafka/waterdrop/blob/master/lib/waterdrop/producer.rb#L69

The moment rdkafka client establishes connection, it is no longer safe to fork the process without recreating the producer.

dezka:
Hi @mensfeld, thank you for the information, glad we're not alone here. This seems to be a good solution.

[...]

PR #200, opened 2022-06-07
fix crash on empty partition key #190

mensfeld:
ref https://github.com/appsignal/rdkafka-ruby/issues/190

For an empty string we use null pointer that is supported.

The failing spec is irrelevant to this case.


leonmaia:
Thanks for the PR @mensfeld. The change LGTM. I'll go ahead and merge this @thijsc.

Issue #168, opened 2021-09-06
Format inconsistency between producing and consuming of messages with headers


mensfeld:
When producing messages, this doc specifies that both keys and values should be string:

https://github.com/appsignal/rdkafka-ruby/blob/b7c790b21427ad419d8144d6c56b0655e2148adc/lib/rdkafka/producer.rb#L81

but when received, they (headers) are casted to symbols:

https://github.com/appsignal/rdkafka-ruby/blob/b7c790b21427ad419d8144d6c56b0655e2148adc/lib/rdkafka/consumer/headers.rb#L54

This creates inconsistency of expectations and may be confusing to developers.

mensfeld:
Yes

PR #207, opened 2022-07-26
Issue #168/string keys for consumer message headers

ColinDKelley:
Addresses issue #168:
- Change `Headers` to return `Hash<String, String>` as documented, starting with corresponding spec change.
- Rdoc updates.

These simplifications are also included:
- Change `Headers` to a `Module` since it is only a namespace and not a `Class`.
- Use `__dir__` for the directory of the current `__FILE__`.

Note: this is a breaking change since any code that is reading the `headers` today with symbols will need to change to strings. Because the version number starts with `0.`, Sem Ver doesn't require a major version bump. But we should advertise it broadly and at least do a minor version bump.

### DIFF
--- a/spec/rdkafka/producer_spec.rb
+++ b/spec/rdkafka/producer_spec.rb
@@ -406,9 +406,9 @@ describe Rdkafka::Producer do
 
     expect(message.payload).to eq "payload headers"
     expect(message.key).to eq "key headers"
-    expect(message.headers[:foo]).to eq "bar"
-    expect(message.headers[:baz]).to eq "foobar"
-    expect(message.headers[:foobar]).to be_nil
+    expect(message.headers['foo']).to eq "bar"
+    expect(message.headers['baz']).to eq "foobar"
+    expect(message.headers['foobar']).to be_nil
   end
 
   it "should produce a message with empty headers" do
--- a/spec/rdkafka/producer_spec.rb
+++ b/spec/rdkafka/producer_spec.rb
@@ -187,7 +187,7 @@ describe Rdkafka::Producer do
     # Close producer
     producer.close
 
-    # Consume message and verify it's content
+    # Consume message and verify its content
     message = wait_for_message(
       topic: "produce_test_topic",
       delivery_report: report,
@@ -211,7 +211,7 @@ describe Rdkafka::Producer do
     )
     report = handle.wait(max_wait_timeout: 5)
 
-    # Consume message and verify it's content
+    # Consume message and verify its content
     message = wait_for_message(
       topic: "produce_test_topic",
       delivery_report: report,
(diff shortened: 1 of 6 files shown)


thijsc (changes requested):
Looks good! There's one spec failing.

spec/rdkafka/producer_spec.rb
```diff
@@ -404,9 +404,9 @@ def call(_, handle)
 
     expect(message.payload).to eq "payload headers"
     expect(message.key).to eq "key headers"
-    expect(message.headers[:foo]).to eq "bar"
-    expect(message.headers[:baz]).to eq "foobar"
-    expect(message.headers[:foobar]).to be_nil
+    expect(message.headers['foo']).to eq "bar"
```
geoff2k, line 407: Nitpick: could we be consistent about using double quotes for strings?

geoff2k (commented):
>  Because the version number starts with 0., Sem Ver doesn't require a major version bump. But we should advertise it broadly and at least do a minor version bump.

Is it perhaps worth discussing (maybe not in this PR), when a major version bump to rdkafka-ruby might be appropriate?

ColinDKelley:
@thijsc
- I fixed that spec.
- Also, I added a spec for `Headers` since that was missing: https://github.com/appsignal/rdkafka-ruby/pull/207/commits/0891e4d3d7c6db4d319a275613d378dcbf5c9f06
- And I was getting really nervous about breaking callers who might still expect Symbol keys. So I tried this pattern that raises an `ArgumentError` if they make that mistake. It's sort of the opposite of HashWIthIndifferentAccess. :) What do you think? https://github.com/appsignal/rdkafka-ruby/pull/207/commits/bfc9eaf763d52abb6e10bea243c67b39afd66fb9

spec/rdkafka/producer_spec.rb
ColinDKelley: @geoff2k Fixed!

ColinDKelley:
@thijsc
Here's a possible interim approach on Symbol-vs-String that is pretty close to `ActiveSupport::HashWithIndifferentAccess` but it's inline/dependency-free. It will tolerate a Symbol key (by converting it to String) but it emits a `Kern.warn` deprecation warning first. This will give developers some notice and a bit of time to switch over without breaking anything. (It documents that when version 1 ships, we will at that point break their code.) https://github.com/appsignal/rdkafka-ruby/pull/207/commits/ef94f01476553f94f2207706180a6c33c47d0351

ColinDKelley:
@thijsc Just following up to see if you have any thoughts on the above?

thijsc:
I think having backwards compatibility for a while would be nice.

ColinDKelley:
@thijsc Will do.

ColinDKelley:
@thijsc The build here is green after the rebase. 👍

mensfeld:
@thijsc there is one problem with this PR. Previous headers were not frozen. While it's not a big deal I want to point out, that this **is** an API change.

PR #209, opened 2022-08-03
bump rdkafka to 1.9.2

mensfeld:
This PR updates rdkafka to 1.9.2 to address offset management issues.


mensfeld:
Nvmd and sorry for the noise :pray:

thijsc:
Why close it?

thijsc:
> to address offset management issues.

Which issues did you see? We're chasing a segfault that seems to be related to offsets in our production environment, I wonder if your issue is the same.

mensfeld:
1.9.2 is a maintenance release: https://github.com/edenhill/librdkafka/releases/tag/v1.9.2

if you find it useful, I can reopen this PR. I just noticed that it would not fix anything.

If you are interested in preventing segfaults you should have pinged me ;) I know how to trigger many and took me a while to track and prevent rdkafka misuses in karafka to stabilize the test suite.

Most of the time they are related to threading or performing two synchronous actions like poll and close when they cannot based on librdkafka docs. There are also segfaults when you try to close the consumer while still holding on to librdkafka objects (all needs to be released prior).

In regards to producer (as previous was on consumer side) most of my old segfaults were related to finalization or again misuses.

mensfeld:
> Which issues did you see? 

Actually none. The issue I've seen came from me not being aware that I was running less than 1.9.0. On 1.9.0 in regards to offsets all is good and none of my integrations fail.

thijsc:
> There are also segfaults when you try to close the consumer while still holding on to librdkafka objects (all needs to be released prior).

Yeah I want to look into fixing these somehow. Maybe we can trick the GC to do the right thing.

mensfeld:
You can try tricking it, properly wrapping the code, or releasing the objects. Edenhill mentioned that few times in the docs, no two non-async operations should ever happen and that all the references to a consumer need to be freed before the consumer is closed. 

Once I tackled the concurrency in both consumer and producer on my side, this problem did not occur in any of my integration tests.

thijsc:
It's working fine for us in production mostly, only sometimes on exit of a process it segfaults. I think a gem should never segfault, even if you're doing it wrong. I'd love to find some way to at least have it raise exceptions if you try to do certain things that could lead to a segfault.

mensfeld:
I agree.

If you want, I can give you a reproduction of this easily. I'm going on a holiday now so won't be able to fix but maybe you will find repro useful.

thijsc:
> If you want, I can give you a reproduction of this easily

Yes please

mensfeld:
First case out of my head:

```
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rdkafka'
end

config = {
  :"bootstrap.servers" => "localhost:9092",
  :"group.id" => "ruby-test"
}

Rdkafka::Config.new(config).admin.create_topic('test')
```

will raise an error and then segfault at least for ruby 3.1.2

mensfeld:
just keep in mind one more thing: any threading non-thread safe operations from librdkafka itself will block ruby completely if you run `.close` the same time:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rdkafka'
end

config = {
  :"bootstrap.servers" => "localhost:9092",
  :"group.id" => "ruby-test",
  'allow.auto.create.topics': true
}

Rdkafka::Config.new(config).admin.create_topic('test', 1, 1)
consumer = Rdkafka::Config.new(config).consumer
consumer.subscribe('test')

sleep 1

Thread.new { loop { consumer.poll(1000) } }

consumer.close
```

this you may have to run few times to make it click. Note that timeout is 1 second but the process will hang forever.

mensfeld:
One more thing I did check: it seems that both  raw librdkafka message and raw tpls are freed correctly. I cannot be 100% certain but I never had any issues in those areas and looked the code and tried to do repros.

thijsc:
I tried the admin create one, with an update because the api is slightly different:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rdkafka'
end

config = {
  :"bootstrap.servers" => "localhost:9092",
  :"group.id" => "ruby-test"
}

Rdkafka::Config.new(config).admin.create_topic('test', 16, 1)
```

It does not raise and error or segfault for me locally.

thijsc:
Do you have another example @mensfeld?

mensfeld:
@thijsc you will have to wait as I am on a PTO :( it also may differ in between mac and linux

In general, all the cases I tackled in karafka and waterdrop were related to running blocking operations in threads while trying to close/stop stuff or not closing properly producer/consumer at exit.

For production, we tackled it together already.

I will be back on Monday and will try to recall more cases.

> It does not raise and error or segfault for me locally.

It still does for me on a ubuntu laptop.

thijsc:
> It still does for me on a ubuntu laptop.

Maybe that's the hint I needed. I've been doing all my testing on Mac so far, this could be a Linux specific issue.

PR #211, opened 2022-08-16
Fix segfault on exit

thijsc:
Work in progress, trying to root out any remaining segfaults.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -16,7 +16,7 @@ blocks:
         - name: bundle exec rspec
           matrix:
             - env_var: RUBY_VERSION
-              values: [ "2.6.10", "2.7.6", "3.0.4", "3.1.2", "jruby-9.3.1.0"]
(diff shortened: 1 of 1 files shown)


mensfeld:
Can we expose the native kafka on admin? Otherwise I need to do tricks like so:

```ruby
Rdkafka::Metadata.new(admin.instance_variable_get('@native_kafka').inner)
```

thijsc:
It's not really something that a normal user would need. Do you need it just for the stability test? Or also from normal code?

mensfeld:
I'm not a normal user ;) I need it for both stability as well as my internal work on the UI for karafka. I can patch it but then it creates issues like the one above :sweat:

mensfeld:
1. Messages production works as expected. Since I support various versions of rdkafka in WaterDrop I run it locally against few and new one behaves the same way.

mensfeld:
2. Messages consumption + admin works as expected. I do not run specs atm against various versions of rdkafka (would eat up a lot of CI but I probably should) but I run my stress suite against 0.13 and no failures and no behavior deviations.

PR #215, opened 2022-09-13
Include topic name in delivery report

maeve:
I want to be able to access the topic name inside a delivery callback, but the delivery report only includes the partition and offset for the delivered message. The underlying dr_msg_cb in librdkafka has access to the message with topic, so this PR introduces changes to pull the topic name from there.


thijsc (approved):
Nice pull!

PR #221, opened 2022-10-03
Support the incremental partition assign and unassign APIs for the cooperative rebalancing protocol when a rebalance listener is set

methodmissing:
## Problem

When setting `partition.assignment.strategy` to `cooperative-sticky`, we observed the following error in logs as a warning (and the consumer made 0 progress as no partitions are assigned):

`W, [2022-10-03T11:02:55.219884 #65446]  WARN -- : rdkafka: [thrd:main]: Group "foo": application *assign() call failed: Changes to the current assignment must be made using incremental_assign() or incremental_unassign()`

Upon further investigation we realised that special API's are [required to be called](https://github.com/edenhill/librdkafka/blob/b871fdabab84b2ea1be3866a2ded4def7e31b006/src-cpp/rdkafkacpp.h#L971-L982):

```
   * For eager/non-cooperative `partition.assignment.strategy` assignors,
   * such as `range` and `roundrobin`, the application must use
   * assign assign() to set and unassign() to clear the entire assignment.
   * For the cooperative assignors, such as `cooperative-sticky`, the
   * application must use incremental_assign() for ERR__ASSIGN_PARTITIONS and
   * incremental_unassign() for ERR__REVOKE_PARTITIONS.
   *
   * Without a rebalance callback this is done automatically by librdkafka
   * but registering a rebalance callback gives the application flexibility
   * in performing other operations along with the assinging/revocation,
   * such as fetching offsets from an alternate location (on assign)
   * or manually committing offsets (on revoke).
```

Those are:

* `rd_kafka_rebalance_protocol` (to query the effective rebalance protocol)
* `rd_kafka_incremental_assign` (to be called when partitions are assigned)
* `rd_kafka_incremental_unassign` (to be called when partitions are revoked)

This ☝️ is true only for when a rebalance callback is registered, otherwise `librdkafka` will do the right thing ™️ 

## The implementation

* Added support for the above APIs (`rd_kafka_rebalance_protocol`, `rd_kafka_incremental_assign` and `rd_kafka_incremental_unassign`)
* Conditionally call the correct assignment APIs on events `RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS` and `RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS` depending on the rebalancing protocol in effect
* Implemented a test case for `partition.assignment.strategy=coperative-sticky` which would just hang otherwise and throw warning `rdkafka: [thrd:main]: Group "foo": application *assign() call failed: Changes to the current assignment must be made using incremental_assign() or incremental_unassign()`

cc @bdbene @kgalieva

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.13.0
+* Supports cooperative sticky partition assignment in the rebalance callback (methodmissing)
+
 # 0.12.0
 * Bumps librdkafka to 1.9.0
 * Fix crash on empty partition key (mensfeld)
--- a/lib/rdkafka/bindings.rb
+++ b/lib/rdkafka/bindings.rb
@@ -112,6 +112,7 @@ module Rdkafka
     attach_function :rd_kafka_conf_set_stats_cb, [:pointer, :stats_cb], :void
     callback :error_cb, [:pointer, :int, :string, :pointer], :void
     attach_function :rd_kafka_conf_set_error_cb, [:pointer, :error_cb], :void
+    attach_function :rd_kafka_rebalance_protocol, [:pointer], :string
 
     # Log queue
     attach_function :rd_kafka_set_log_queue, [:pointer, :pointer], :void
@@ -175,6 +176,8 @@ module Rdkafka
     attach_function :rd_kafka_unsubscribe, [:pointer], :int
     attach_function :rd_kafka_subscription, [:pointer, :pointer], :int
     attach_function :rd_kafka_assign, [:pointer, :pointer], :int
+    attach_function :rd_kafka_incremental_assign, [:pointer, :pointer], :int
+    attach_function :rd_kafka_incremental_unassign, [:pointer, :pointer], :int
     attach_function :rd_kafka_assignment, [:pointer, :pointer], :int
     attach_function :rd_kafka_committed, [:pointer, :pointer, :int], :int
     attach_function :rd_kafka_commit, [:pointer, :pointer, :bool], :int, blocking: true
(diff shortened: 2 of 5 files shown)


mensfeld:
This implementation does not work as expected (probably fault is not here though as it is "by the book"). I've been trying to make run exactly the same code myself and to make it work with my integration suite (one actually running against kafka) and once in a while I do encounter this issue: https://github.com/edenhill/librdkafka/issues/3891 + some incidents when partitions that are assigned back after a series of rebalances do not pick up for processing.

methodmissing:
Thanks Maciej, any particular Kafka version, or across many?

mensfeld:
I checked few versions released by wurstmeister via docker. Also if we are talking about a final solution (if I may), maybe it would be good to separate the callback into two and inject the correct one upon initialization. Though that also may be a different task. 

The most common case that I've seen that causes my specs to hang is a case where upon leaving of a group the remaining work is not reassigned back despite receiving the appropriate events on the rest of the consumers in the group. I am still debugging this...

mensfeld:
@methodmissing I think I figured out the issue. All the consumer instances of librdkafka in one process need to either have the rebalance callback or not. It cannot be one vs the other. Please give me 1-2 days to confirm that and convert into integration specs.

methodmissing:
This is interesting, consumer instances should however be decoupled / segregated even within the same process space.
> All the consumer instances of librdkafka in one process need to either have the rebalance callback or not.

mensfeld:
Yeah, was able to confirm that once the manual commit during rebalance is removed things work as expected:

https://github.com/karafka/karafka/blob/29b6b59fefcf49726497e0811a156ec7882d8c87/spec/integrations/rebalancing/cooperative_sticky.rb#L63

so this PR technically is ok, though I still think it should be documented somewhere for the iterator-based consumer that is part of rdkafka API.

methodmissing:
@thijsc how would you like to proceed here?

> so this PR technically is ok, though I still think it should be documented somewhere for the iterator-based consumer that is part of rdkafka API.

thijsc:
Could you rebase on main? A fix was merged that will hopefully lead to a green build for this.

methodmissing:
@thijsc rebased, failed on the `zstd` issue on the Ruby 3.1.2 build 🤷‍♂️

thijsc:
The build is actually green, that spec is intentionally pending. There’s the elusive segmentation fault all the way at the end that’s the issue.

methodmissing:
@thijsc possible to merge this one down as well? I'm concurrently looking into the generalized segfault on shutdown in my Linux environment and it's been slow to repro today

mensfeld:
I can also confirm this solution works with my specs ;)

mensfeld:
https://my.diffend.io/gems/rdkafka/0.13.0.beta.1/0.13.0.beta.2

works.

PR #224, opened 2022-10-26
Remove consumer argument in rebalance hooks

thijsc:
We were creating a new consumer, which is not the same as the original object. If access to the consumer is needed that reference needs to be stored in the opaque.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Fix leaking FFI resources in specs (mensfeld)
 * Improve specs stability (mensfeld)
 * Make metadata request timeout configurable (mensfeld)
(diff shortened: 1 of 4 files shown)


thijsc:
Looks like there's still a bug in this one.

mensfeld:
@thijsc fixed.

thijsc:
Cool. thanks

mensfeld:
@thijsc just to be sure, maybe one more beta with this? :D so I can run my integrations?

thijsc:
Pushed!

mensfeld:
Thanks. I will run the final set of integration tests on it somewhere around RubyKaigi.

PR #225, opened 2022-10-27
Remove blocking flag for poll and close

thijsc:
Poll and close may never be called at the same time. Remove the blocking flag in the bindings so the GIL will help us prevent this.


thijsc:
This breaks some stuff. We'd be better off with a mutex per client probably.

PR #227, opened 2022-11-08
Lock access to the native kafka client

thijsc:
Lock access from the admin and consumer class. This prevents rdkafka from segfauling if the native instance is accessed during close.

### DIFF
--- a/lib/rdkafka/bindings.rb
+++ b/lib/rdkafka/bindings.rb
@@ -41,9 +41,9 @@ module Rdkafka
 
     # Metadata
 
-    attach_function :rd_kafka_memberid, [:pointer], :string
-    attach_function :rd_kafka_clusterid, [:pointer], :string
-    attach_function :rd_kafka_metadata, [:pointer, :int, :pointer, :pointer, :int], :int
+    attach_function :rd_kafka_memberid, [:pointer], :string, blocking: true
+    attach_function :rd_kafka_clusterid, [:pointer], :string, blocking: true
+    attach_function :rd_kafka_metadata, [:pointer, :int, :pointer, :pointer, :int], :int, blocking: true
     attach_function :rd_kafka_metadata_destroy, [:pointer], :void
 
     # Message struct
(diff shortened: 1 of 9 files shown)


thijsc:
> Segmentation fault (core dumped)

Well this didn't fix it. Trying with a locked producer too.

thijsc:
I've been working on this some more and the current incarnation seems to work. @mensfeld wdy?

lib/rdkafka/bindings.rb
```diff
@@ -170,27 +170,26 @@ class TopicPartitionList < FFI::Struct
...
     attach_function :rd_kafka_consumer_poll, [:pointer, :int], :pointer, blocking: true
     attach_function :rd_kafka_consumer_close, [:pointer], :void, blocking: true
-    attach_function :rd_kafka_offset_store, [:pointer, :int32, :int64], :int
-    attach_function :rd_kafka_pause_partitions, [:pointer, :pointer], :int
-    attach_function :rd_kafka_resume_partitions, [:pointer, :pointer], :int
-    attach_function :rd_kafka_seek, [:pointer, :int32, :int64, :int], :int
+    attach_function :rd_kafka_offset_store, [:pointer, :int32, :int64], :int, blocking: true
+    attach_function :rd_kafka_pause_partitions, [:pointer, :pointer], :int, blocking: true
+    attach_function :rd_kafka_resume_partitions, [:pointer, :pointer], :int, blocking: true
+    attach_function :rd_kafka_seek, [:pointer, :int32, :int64, :int], :int, blocking: true
```
geoff2k, line 192: Should `rd_kafka_conf_set_rebalance_cb` also be set to blocking and get the `with_inner` treatment?

mensfeld: Just for the record and to be sure everyone here knows: blocking releases GIL/GVL.

lib/rdkafka/native_kafka.rb
```diff
@@ -51,11 +67,19 @@ def close(object_id=nil)
...
+      Thread.new do
+        # Lock the mutexes for polling and method access
+        # to make sure we are not destroying during an
+        # ongoing operation
+        @method_mutex.lock
+        @poll_mutex.lock
+
+        # Destroy the client
+        Rdkafka::Bindings.rd_kafka_destroy(@inner)
+      end.join
```
mensfeld, line 81: can we add ensure here to unlock? I feel bad leaving this dangling

mensfeld (approved):

mensfeld:
On first glimps looks good. My only worry here is that the kafka inner API (which is totally unofficial but whatever ;) ) is changing. On the other hand it is changing for better so :shrug: 

There is one more case that hangs: try to create a topic when Kafka cluster is down. Admin will hang forever.

thijsc:
I have one more idea to maybe do this a bit nicer, going to try that.

> My only worry here is that the kafka inner API (which is totally unofficial but whatever ;) ) is changing.

Yeah that's a private API :-). But the with inner method should be stable after this is merged.

> There is one more case that hangs: try to create a topic when Kafka cluster is down. Admin will hang forever.

I will take a look at this.

lib/rdkafka/bindings.rb
thijsc: > Just for the record and to be sure everyone here knows: blocking releases GIL/GVL.

👍 I think this should work well now, since we won't lock the GIL, but there's still a lock on the consumer around all access. This way you can have multiple consumers in multiple threads that don't block each other, while still making sure the method calls to the C codes are sync.

mensfeld:
@thijsc 

> I will take a look at this.

Plese don't release 0.13 after this as I want to add one more thing: metadata cache for partitions. I already contacted confluent devs on risks of doing that and they confirmed it's a good approach.

This should be delivered tops next week. It's a high priority for me.

thijsc:
I found a still nicer way to do it: 10e3183f6aec37688e7e622924550ff009627cd7

lib/rdkafka/native_kafka.rb
thijsc: Implemented in a different way now

mensfeld:
@thijsc so should I review it? :D

thijsc:
Yeah I think this should be the, one rebasing it to one commit.

mensfeld:
@thijsc I see you took an inspiration (great!) https://github.com/karafka/karafka/blob/master/lib/karafka/admin.rb#L125 ;)

mensfeld (approved):

mensfeld:
Looks good. Once you finalize things ping me and we can run it against the karafkas integration test suite to check the stability.

thijsc:
Even with this approach there are ocassional segfaults. I think there still needs to be some extra check in librdkafka itself to 100% fix this. I think there's nothing left on the Ruby side now. I'm going to merge this.

methodmissing:
@thijsc thoughts on cutting a new release with these changes?

thijsc:
We should. There is a beta out, could you give it a spin? I haven't heard back from people yet.

mensfeld:
You did from me. It works well with Karafka. There's one PR I want to add but can do after the release and will go with the next one.

methodmissing:
We've been on `0.13.0.beta.3` since it was cut, but hungry for more :smile:

https://media.giphy.com/media/lrc1TZHRYxj7lGM3Vg/giphy.gif

methodmissing:
Also folding https://github.com/appsignal/rdkafka-ruby/pull/229 into a beta release could be interesting, given it's an additive API and feature.

thijsc:
I've pushed one more gem because the latest commit wasn't in there yet. I'll push out 0.13 soon.

After that I want to push one getting rid of the now-rare segfault once more. And do a round of stuff that could be a breaking change. And then I'd like to cut a 1.0, I think we're ready for that.

mensfeld:
@thijsc if you give me one day I can rerun all my regressions against new beta.

mensfeld:
```
Segmentation fault (core dumped)
```

:sob: 

investigating...

mensfeld:
Does not work too well: https://github.com/karafka/waterdrop/actions/runs/3882428479/jobs/6622575727

I will investigate but I do see frequent crashes.

mensfeld:
Few things:
1. It works until you end up not closing producer
2. Something breaks when finalizer is shutting down producers forcefully (without user closing it)

I think that due to severity (70% of times crashes) I would recommend postponing the release until we figure out what is happening. The code will work for long running processes but any case where producer is not closed may cause this.

thijsc:
I had this idea that doing the cleanup in that poll would be most reliable, because then everything is sync by nature. But maybe the previous approach is better :-).

mensfeld:
> maybe the previous approach is better :-).

well this is what my specs say.

PR #229, opened 2022-11-16
Add offsets_for_times method on consumer

timflapper:
Expose `offsets_for_times` method on the Consumer class.

This method does a lookup of offsets based on timestamps for each partition in the TopicPartitionList and returns a new TopicPartitionList with the `offsets` set the the offset of the first message in said partition whose timestamp is equal to or higher than the timestamp provided. The partitions in the TopicPartitionList do not need to be assigned to the consumer.

This PR exposes `raw_timestamp` as part of the `Message` class. It's used by the test to make it easier to retrieve the message's raw timestamp. For anybody wanting to use `offsets_for_times` it's convenient to have direct access instead of having to convert the `Time` object in `timestamp` back to a millisecond timestamp integer.


lib/rdkafka/consumer.rb
```diff
@@ -385,6 +385,37 @@ def seek(message)
...
+      if !list.is_a?(TopicPartitionList)
+        raise TypeError.new("list has to be a TopicPartitionList")
+      end
+
+      tpl = list.to_native_tpl
+
+      response = Rdkafka::Bindings.rd_kafka_offsets_for_times(
+        @native_kafka.inner,
+        tpl,
+        -1 # timeout
```
mensfeld, line 407: not having timeout may create hard to debug issues. I would opt on providing it even if high (60 seconds for example).

lib/rdkafka/consumer.rb
```diff
@@ -385,6 +385,37 @@ def seek(message)
       end
     end
 
+    # Lookup offset for the given partitions by timestamp.
+    #
+    # @ param list [TopicPartitionList] The TopicPartitionList with timestamps instead of offets
```
mensfeld, line 390: typo `@param`

lib/rdkafka/consumer/message.rb
```diff
@@ -28,6 +28,10 @@ class Message
       # @return [Time, nil]
       attr_reader :timestamp
 
+      # This message's raw timestamp, if provided by the broker
+      # @return [Time, nil]
+      attr_reader :raw_timestamp
```
mensfeld, line 33: I have a mixed feeling about exposing this. How often does the messages timestamp lookup occurs? That is: does frequency of re-computing the raw time value justifies expanding the API.

mensfeld (changes requested):

lib/rdkafka/consumer/message.rb
timflapper: I'm happy to remove this change from the PR but it might be worth considering the use case of wanting access to the original data as-is. As a wrapper library it makes sense to expose the raw data as much as possible.

mensfeld: @timflapper I don't have a super strong opinion. I just don't like expanding APIs unless needed because then you create a long-lasting contract with the users ;)

@thijsc should probably make a decision here

thijsc: I think it would be more Ruby-esque to use times in the input topic partition list. We're trying to make this as nice to use as possible. We could add a field for this on the Ruby TPL and convert it when calling librdkafka.

timflapper: Ok. I'll see if I can get it to work to allow the TPL to detect offset as a Time object to convert to a millisecond timestamp.

My personal note on this is that I would personally choose to keep timestamps in their original form (perhaps naming them timestamp_ms to notify the timestamp format) and letting the library's user convert it to a Time object if they wish. The time conversion means that the library can never be as fast as it can possibly be since every message has a data transformation that may or may not be useful to the user but that does slow down the message processing, even if very slightly.

mensfeld: @timflapper this would change the contract for this package. It could however be lazy evaluated for sure. Nonetheless IMHO worth opening a discussion about it outside of scope of this task.

lib/rdkafka/consumer.rb
```diff
@@ -385,6 +385,37 @@ def seek(message)
       end
     end
 
+    # Lookup offset for the given partitions by timestamp.
+    #
+    # @param list [TopicPartitionList] The TopicPartitionList with timestamps instead of offets
```
mensfeld, line 390: offsets (typo)

mensfeld (approved):
there's a small typo but aside lgtm

lib/rdkafka/consumer.rb
timflapper: Thanks @mensfeld. Fixed the typo.

lib/rdkafka/consumer/message.rb
thijsc: @timflapper We made a design choice to have this library get as close to idiomatic Ruby as we can. That means that we sometimes sacrifice some performance.

thijsc: I guess you could do the conversion in a lazy way btw? Often it won't be needed at all.

mensfeld: @thijsc you mean introducing a lazy convertion to the message object? Sounds like a code smell. Imho that would be an unexpected mutability of a simple data object (so far at this point).

thijsc: Why not? We we can store the raw number in a private field and convert it whenever we need. Why is that a code smell?

mensfeld: For me, the message then is no longer just a "close to being frozen" data structure, and it becomes mutable. From the API standpoint, this approach is **nothing** wrong.

lib/rdkafka/consumer.rb
mensfeld: Thanks. I will get back to this PR in the upcoming weeks.

mensfeld:
@timflapper would you mind resolving conflicts with the current master? Sorry for the delay :pray:

timflapper:
@mensfeld, apologies. I missed this one. The conflict should be resolved now.

mensfeld:
Thanks @timflapper I will pick it up from here. No worries. I have it working locally already. I don't want to target it to the current beta (when it stops being beta) but I will merge into the next one as a new functionality. The reason is simple: there are already enough big changes in the current beta and would not want to introduce another

mensfeld:
I'll align it with the recent changes to the native locks.

mensfeld:
@timflapper not to keep in on you any longer, retargeting to `add-offsets_for_times-method-on-consumer`, merging there and will apply alignments.

PR #230, opened 2022-11-18
Check metadata cache for topic's partition count if available

ttstarck:
### Problem
Currently when a message is produced without specifying a partition, a metadata request gets made to the kafka broker to get the number of partitions for the topic. That means every time we produce a message without specifying the partition, we have to make a request for the metadata. This is a lot of unnecessary requests happening to get a value (topic partition count) that changes _very_ infrequently, if ever.

### Solution
In this pull request, I have made changes to how the partition count for a topic is found, such that the we first search the metadata cache for the topic's partition count. If the topic exists in the metadata cache, then we use that partition count and do not make a request for metadata. If the topic doesn't exist in the metadata cache, then we make a request to the broker for the metadata (how it currently works).

Librdkafka source code: [rd_kafka_metadata_cache_topic_get](https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_metadata_cache.c#L635).

### DIFF
--- a/lib/rdkafka/bindings.rb
+++ b/lib/rdkafka/bindings.rb
@@ -45,6 +45,7 @@ module Rdkafka
     attach_function :rd_kafka_clusterid, [:pointer], :string
     attach_function :rd_kafka_metadata, [:pointer, :int, :pointer, :pointer, :int], :int
     attach_function :rd_kafka_metadata_destroy, [:pointer], :void
+    attach_function :rd_kafka_metadata_cache_topic_get, [:pointer, :string, :int], :pointer
 
     # Message struct
 
--- a/lib/rdkafka/metadata.rb
+++ b/lib/rdkafka/metadata.rb
@@ -4,6 +4,37 @@ module Rdkafka
   class Metadata
     attr_reader :brokers, :topics
 
+    class << self
+      # Topic metadata pulled from the metadata cache for a given topic.
+      # Note: Will return nil in the following scenarios:
+      #   1. Metadata has not been initialized.
+      #   2. Topic could not be found in the cache.
+      #
+      # @param native_client [FFI::Pointer] Pointer to the native kafka client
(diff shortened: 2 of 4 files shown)


mensfeld:
```
Function 'rd_kafka_metadata_cache_topic_get' not found in [/home/semaphore/rdkafka-ruby/lib/rdkafka/../../ext/librdkafka.so]
```

I wonder why :thinking:

ttstarck:
Ahh... It looks like `rd_kafka_metadata_cache_topic_get` is not apart of the public API. However it does seem to be available when compiling librdkafka on debian and macos, but seemingly not ubuntu.

I'll close this PR as this solution will most likely not work. However, it still leaves the main problem open that we are making a request for `metadata` every time we produce a message without specifying a partition. I'll look into creating an issue to track this.

mensfeld:
> Currently when a message is produced without specifying a partition, a metadata request gets made to the kafka broker to get the number of partitions for the topic.

Are you are referring to this code? https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/producer.rb#L107

mensfeld:
@ttstarck I contacted confluent about caching this value and they said it's ok to cache. I implemented a patch in waterdrop and will backport it to rdkafka when I have some time.

ttstarck:
> Are you are referring to this code? https://github.com/appsignal/rdkafka-ruby/blob/main/lib/rdkafka/producer.rb#L107

Yup, exactly!

>I contacted confluent about caching this value and they said it's ok to cache. I implemented a patch in waterdrop and will backport it to rdkafka when I have some time.

Awesome! Thank you!

ttstarck:
CC @mensfeld I've opened a new PR pulling in those changes you've made in karafka-rdkafka

https://github.com/appsignal/rdkafka-ruby/pull/282

PR #237, opened 2023-01-31
Add continued support for `v0.10.x` releases (update `LIBRDKAFKA` source sha256).

drn:
For a variety of unfortunate reasons, one of our codebases is using ruby `< 2.6`, so we're stuck on `v0.10.x` of this gem. Given it looks like the `LIBRDKAFKA` source SHA256 checksum has changed, we can no longer install the `v0.10.0` version of `rdkafka-ruby`. Of course we could upgrade to a newer version of the gem, but that'd require us to upgrade ruby (which we're regrettably blocked on a few items on).

It'd be very helpful if someone could cut a `v0.10.1` release of the gem with the SHA update and unblock our deployment process.

Thank you!


drn:
Unfortunately, it looks like there is no tag on `v0.10.0` (https://github.com/appsignal/rdkafka-ruby/commit/90285c9ae8741e4a2b7a478cdba783f783b7a65c) so I was unable to open this PR against that SHA. I think all that would need to happen to unblock us would be to cut a release with just this commit on top of the https://github.com/appsignal/rdkafka-ruby/commit/90285c9ae8741e4a2b7a478cdba783f783b7a65c. Thanks in advance!

drn:
Very strange. It looks like the SHA switched back and installation is now working

This was the original issue on bundle install:
```
Downloaded file 'ports/archives/v1.5.0' has wrong hash: expected: f7fee59fdbf1286ec23ef0b35b2dfb41031c8727c90ced6435b8cf576f23a656 is: 3fbbfb19eb859b845854855491f93cfb9b59ee13f78edf5824204025753aebee
```

It looks like this has since resolved itself. Closing now

mensfeld:
@drn GH reverted the change. They broke checksums for releases.

drn:
Ahh, that would explain the other issues we were having today. Thanks @mensfeld

Issue #234, opened 2023-01-10
0.13.0.beta4 crash for not closed producers


mensfeld:
New issue for myself so we keep track during investigation:

1. It works until you end up not closing producer
2. Something breaks when finalizer is shutting down producers forcefully (without user closing it)

I think that due to severity (70% of times crashes) I would recommend postponing the release until we figure out what is happening. The code will work for long running processes but any case where producer is not closed may cause this.

ref https://github.com/karafka/waterdrop/actions/runs/3882428479/jobs/6622575727

mensfeld:
Undefining `ObjectSpace.define_finalizer(...` for producer and using the one from WaterDrop seems to fix this issue.

Doing the other way around (removing the WaterDrop one and relying on rdkafka one) makes it less severe but still existing.

mensfeld:
There is a hang on ruby exit, which indicates, that polling happens. This leads me to an educated guess, that some librdkafka resources are no longer available after poll to destroy, hence the crash.

mensfeld:
Yeah: once destroy happens in the close method, no seg faults for the producer.

mensfeld:
When below patch is pushed:

```ruby
module Rdkafka
  # @private
  # A wrapper around a native kafka that handles all access
  class NativeKafka
    def initialize(inner)
      @inner = inner
      @mutex = Mutex.new

      # Start thread to poll client for delivery callbacks
      @polling_thread = Thread.new do
        loop do
          Rdkafka::Bindings.rd_kafka_poll(@inner, 250)
          # Destroy consumer and exit thread if closing and the poll queue is empty
          break if Thread.current[:closing] && Rdkafka::Bindings.rd_kafka_outq_len(@inner) == 0
        end
      end

      @polling_thread.abort_on_exception = true
      @polling_thread[:closing] = false

      @closing = false
    end

    def close(object_id=nil)
      return if closed?

      # Indicate to the outside world that we are closing
      @closing = true

      # Indicate to polling thread that we're closing
      @polling_thread[:closing] = true

      @mutex.synchronize do
        return unless @inner

        # Wait for the polling thread to finish up
        @polling_thread.join

        Rdkafka::Bindings.rd_kafka_destroy(@inner)
        @inner = nil
      end
    end
  end
end
```

works as expected:

https://github.com/karafka/waterdrop/actions/runs/3883397933/jobs/6624662007

mensfeld:
[...]

PR #238, opened 2023-01-31
Lock access to native kafka

thijsc:
Put all access on the native kafka behind locks, so that we can exit cleanly.

### DIFF
--- a/spec/rdkafka/consumer_spec.rb
+++ b/spec/rdkafka/consumer_spec.rb
@@ -738,7 +738,7 @@ describe Rdkafka::Consumer do
       #
       # This is, in effect, an integration test and the subsequent specs are
       # unit tests.
-      admin = dkafka_config.admin
+      admin = rdkafka_config.admin
       create_topic_handle = admin.create_topic(topic_name, 1, 1)
       create_topic_handle.wait(max_wait_timeout: 15.0)
       consumer.subscribe(topic_name)
--- a/spec/rdkafka/consumer_spec.rb
+++ b/spec/rdkafka/consumer_spec.rb
@@ -738,7 +738,8 @@ describe Rdkafka::Consumer do
       #
       # This is, in effect, an integration test and the subsequent specs are
       # unit tests.
-      create_topic_handle = rdkafka_config.admin.create_topic(topic_name, 1, 1)
+      admin = dkafka_config.admin
+      create_topic_handle = admin.create_topic(topic_name, 1, 1)
       create_topic_handle.wait(max_wait_timeout: 15.0)
       consumer.subscribe(topic_name)
       produce_n 42
@@ -751,6 +752,7 @@ describe Rdkafka::Consumer do
       expect(all_yields.flatten.size).to eq 42
       expect(all_yields.size).to be > 4
       expect(all_yields.flatten.map(&:key)).to eq (0..41).map { |x| x.to_s }
+      admin.close
     end
 
     it "should batch poll results and yield arrays of messages" do
@@ -793,13 +795,15 @@ describe Rdkafka::Consumer do
     end
 
     it "should yield [] if nothing is received before the timeout" do
-      create_topic_handle = rdkafka_config.admin.create_topic(topic_name, 1, 1)
+      admin = rdkafka_config.admin
+      create_topic_handle = admin.create_topic(topic_name, 1, 1)
       create_topic_handle.wait(max_wait_timeout: 15.0)
       consumer.subscribe(topic_name)
       consumer.each_batch do |batch|
         expect(batch).to eq([])
         break
       end
+      admin.close
     end
 
     it "should yield batchs of max_items in size if messages are already fetched" do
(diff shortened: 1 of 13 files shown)


mensfeld:
ref https://github.com/appsignal/rdkafka-ruby/issues/234

you want me to crash-test it? :rofl:

thijsc:
> you want me to crash-test it? 🤣

Yes please :-)

mensfeld:
@thijsc FYI did not forget about it. Should be able to do this at the end of this week for both waterdrop and karafka

mensfeld:
@thijsc one more thing that is worrying (not related to this code but may be worth fixing altogether). Sometimes admin actions callback is not called back :/ no idea yet why it's happening, though.

mensfeld:
Sorry, I was occupied running this instead first: https://github.com/appsignal/rdkafka-ruby/pull/242 - for this I can confirm that producer, consumer and admin work as expected without any issues.

For this (238) I will be running this today and tomorrow and will post you.

mensfeld:
Ok running stress tests on producer now...

mensfeld:
@thijsc good news and bad news. Let me sum it up with a meme:

indeks

mensfeld:
gdb claims it hangs on close:

```
from /mnt/software/Karafka/rdkafka-ruby/lib/rdkafka/producer.rb:45:in `close'
from /mnt/software/Karafka/rdkafka-ruby/lib/rdkafka/native_kafka.rb:72:in `close'
from /mnt/software/Karafka/rdkafka-ruby/lib/rdkafka/native_kafka.rb:72:in `join'
```

for not-closed producers.

mensfeld:
ok, giving it a spin from the producer pespective...

mensfeld:
@thijsc question / remark:

The newly introduced `#flush` has a comment like this:

# Call this before closing a producer to ensure delivery of all messages.

but this contradicts the librdkafka docs themselves and usage of `rd_kafka_outq_len`:

 > @brief Wait until all outstanding produce requests, et.al, are completed.
 >        This should typically be done prior to destroying a producer instance
 >        to make sure all queued and in-flight produce requests are completed
 >        before terminating.

Using both is a huge problem for two reasons:
- you cannot flush from a closed producer after the close
- if you flush before the close (including states) there is a small gap where new messages can be dispatched despite the flush because producer is not technically closed yet. This will be a problem in a multi-thread env.

So we can do two things:
- ditch `rd_kafka_outq_len` (not recommended)
- update flush docs to indicate, that close producer flush happens and that this is NOT needed (because it is not) but CAN be used for periodic flushes to dispatch messages faster, et (recommended)

WDYT? This is a big **blocker** for me, because it's either one or the other.

mensfeld:
Actually my PR that got merged could then be reverted if we would document that flush has a potential race condition on close but it is irrelevant because of `rd_kafka_outq_len` usage.

mensfeld:
With my PR reverted AND flush before initialization of close + `rd_kafka_outq_len` on wait, producer is **stable**. No crashes there anymore.

Now for the consumer...

mensfeld:
@thijsc I don't have rights to restart semaphore buuuuut it's :green_heart: 

What I did:
- reorder the shutdown process to ensure that queue is empty and closed before we destroy the inner kafka
- reorder to prevent deadlock
- move kafka start in docker higher up so it fully starts while things are bundled and compiled - I had the same problem with my regressions. Unless Kafka starts and gets a bit of time to do housekeeping, fast topic creation can hang it during warmup.

- My integration producer specs now fully pass under load
- My integration suite also. No segfaults.

nijikon (approved):
LGTM 🚀

mensfeld:
Ok one more challenge I see now for finalizers: trap context. Will fix.

thijsc:
You're on a roll!

mensfeld:
In total there were 14 dangling rdkafka objects, some with finalizers and some without (direct references of inner).

A potential multi-thread race condition was also fixed.

Another crash was caused by leaked consumes being present during fork and attempt to close them when fork was finishing. COW won't work with open sockets.

spec/rdkafka/consumer_spec.rb
```diff
@@ -738,7 +738,8 @@ def new_message
       #
       # This is, in effect, an integration test and the subsequent specs are
       # unit tests.
-      create_topic_handle = rdkafka_config.admin.create_topic(topic_name, 1, 1)
+      admin = dkafka_config.admin
```
nijikon, line 741: `rdkafka_config`

mensfeld: Fixed. Thanks

mensfeld:
Sorry for the noise but I want to trigger the CI at least 5 times in a row to make sure no other things go unnoticed. The fixes to specs explain why CI would crash but the operations would work as expected.

mensfeld:
8/8 green after the last fixes. @thijsc feel free to take a look. This should conclude this PR. The stability on the Karafka side is unchanged (100% of things pass for producers and consumers).

At the moment ALL non-closeable objects that were leaking are closed correctly. Nothing that would not be finalized leaks to finalization or forks.

mensfeld (approved):

nijikon (approved):
@mensfeld great work on this! @thijsc you think that this can be merged now?

nijikon:
@thijsc ping

PR #239, opened 2023-01-31
Make metadata timeout configurable

mensfeld:
Similar to other places where `timeout_ms` is used. Just lets make it configurable.


nijikon:
@thijsc can we merge this?

PR #243, opened 2023-02-14
Add create_acl method

johnernaut:
This PR adds functionality which lets a user create ACL's.  It extends the existing `Admin` namespace and has an API that looks like the following:

```ruby
  admin = Rdkafka::Config.new(config).admin

  acl_options = {
    resource_type: :topic,
    pattern_type: :literal,
    acl_operation: :all,
    permission_type: :allow
  }
  create_acl_handle = admin.create_acl('some_name', 'User:*', '*', acl_options)
  create_acl_handle.wait
```

If the approach is agreeable, I'd be happy to work on adding describe/destroy methods for ACL's as well.

Notes:
* This required a re-work of the `docker-compose` file in order to utilize SASL/SCRAM auth which is required (taken from [segmentio/kafka-go](https://github.com/segmentio/kafka-go/blob/main/docker-compose.yml)).  I was unable to come up with the correct config options using the confluent images but I can revisit if it's a blocker.


mensfeld:
I would target it after the native lock release. @johnernaut I am aware of this one. It was not abandoned.

mensfeld:
@johnernaut superseeded by: https://github.com/karafka/rdkafka-ruby/pull/347 (please correct me if I'm wrong and will reopen and take care of it)

PR #247, opened 2023-03-27
Ported to Windows with MSYS2

jake-007:
Windows with MSYS2 has _librdkafka_ available through `pacman` MSYS2 package management. This PR modifies **rdkafka-ruby** to detect it's running on windows with `Gem.win_platform?` and if that's the case, relies on [Rubyinstaller2 MSYS2 Library Dependency](https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers#msys2-library-dependency) to check and install the _librdkafka_ package.

The platform check in the forking spec is replaced as recommended in [Kernel.fork](https://ruby-doc.org/3.1.3/Kernel.html#method-i-fork) Ruby official documentation.

When _librdkafka_ is installed by `pacman` in the MSYS2 system, `ffli_lib` finds it automatically.

### DIFF
--- a/lib/rdkafka/bindings.rb
+++ b/lib/rdkafka/bindings.rb
@@ -17,8 +17,12 @@ module Rdkafka
       end
     end
 
-    ffi_lib File.join(__dir__, "../../ext/librdkafka.#{lib_extension}")
-
+    if Gem.win_platform?
+      ffi_lib 'librdkafka'
+    else
+      ffi_lib File.join(__dir__, "../../ext/librdkafka.#{lib_extension}")
+    end
+      
     RD_KAFKA_RESP_ERR__ASSIGN_PARTITIONS = -175
     RD_KAFKA_RESP_ERR__REVOKE_PARTITIONS = -174
     RD_KAFKA_RESP_ERR__NOENT = -156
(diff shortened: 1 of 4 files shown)


mensfeld:
The problem with this solutions, is that we rely on an external source with external version that is not locked. Any way to lock it to a particular one on windows?

jake-007:
We can specify a version in rdkafka.gemspec as follows:

```ruby
gem.metadata['msys2_mingw_dependencies'] = 'librdkafka=2.0.2' if Gem.win_platform?
```
Though the [Rubyinstaller2 documentation](https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers#msys2-library-dependency) says:

> However keep in mind that MSYS2 usually provides only one version at the same time for each package.

and that version right now is 2.0.2. There are a few recent versions available as files for download in the [MSYS2 repo](https://repo.msys2.org/mingw/x86_64/) but I know of no way to ask _pacman_ to install one.

jake-007:
We can specify a version in rdkafka.gemspec as follows:

```ruby
gem.metadata['msys2_mingw_dependencies'] = 'librdkafka=2.0.2' if Gem.win_platform?
```
Though the [Rubyinstaller2 documentation](https://github.com/oneclick/rubyinstaller2/wiki/For-gem-developers#msys2-library-dependency) says:

> However keep in mind that MSYS2 usually provides only one version at the same time for each package.

and that version right now is 2.0.2. There are a few recent versions available as files for download in the [MSYS2 repo](https://repo.msys2.org/mingw/x86_64/) but I know of no way to ask _pacman_ to install one.

mensfeld:
I still don't know how to tackle this :thinking:

mensfeld:
Closing. Not going to support windows. At least not me

Issue #260, opened 2023-06-16
Reconsider hard locking in favour of flow control


mensfeld:
@thijsc there seems to be one imho critical problem with 0.13.0 approach to locking that I discovered:

because we lock poll, that can take significant amount of time, any operations that could be performed during that time that are fully safe from librdkafka perspective (pausing, seeking, etc) will be locked for an extensive period of time equal at most the polling wait time.

This can impact performance drastically upon cases like this. I think we should reconsider having the long running operations under more granular locking flow.

mensfeld:
Actually now that I start to revisit the librdkafka docs, most of the methods are thread safe. The only thing that needs to be ensured, is that destruction (or few others) are not running in parallel. It would be much better instead of locking to keep track of usage of those (checkpoint into an array or something) and guard. That way there should be no problem with thread-safe but we should not lock (as long as we guard only the checks and not the commands)

mensfeld:
Here's the pseudocode to do so:

```ruby
    def with_inner
      return if @inner.nil?

      caller_name = caller[0]
      @ops_in_progress << caller_name

      result = yield @inner

      p @ops_in_progress
      @ops_in_progress.delete(caller_name)

      result
    end
```

and we just need to ensure, that destroy ain't happening at the time when there's anything running

mensfeld:
ofc caller is fairly slow, we could just use atomic int (or int via mutex) to indicate the progress without the division and guard only those that need to have 0. For example if you have poll in progress close will have to wait until the lock is released but we should start and release lock ONLY for things that collide. As far as I can see the docs, that's only closing stuff.

mensfeld:
So my proposition here is as followed:

1. synchronized thread-safe counter on inner usage
2. a switch for `close` that will ensure that when closing, any new potentially race conditioning async methods go into a mutex
3. same mutex on close to wait for anything that is still running
4. wait on count getting to 0 and lock mutex
5. close

That way we ensure that all operations can go async BUT when we starting closing, we get into it with mutex and that no other ops are running. This will ensure, that nothing collides with `#close` and destroy and also allows to run async operations until close without heavy locking.

thijsc:
Sounds great!

mensfeld:
Note to self: if we expose the mutex from the native client creator to the native client, we can ensure this easily:

```ruby
def with_native
  external_mutex.synchronize { counter up }
  yield(native)
  external_mutex.synchronize { counter down }
end
```

and from the end consumer

[...]

PR #264, opened 2023-06-21
Lock native only on closing and track usage

mensfeld:
This PR:

- intoduces a new native kafka method called `#synchronize` that yields only when no parallel requests are running (including started before the lock), ensuring that a given C binding operation can be started and finished fully without any other.
- all the other potential operations are blocked on the increment mutex
- no need to have mutexed decrement as the zero value reachability is anyhow guarded by sleep
- spec was added to ensure, that long polling is finished before close happens.

Thanks to this approach we ensure that both closing the native as well as any operations (like consumer shutdown) can operate only when no other operations are running (mainly poll that is long running) and that all the resources are cleaned prior.

Since we may have a potential race condition where the client is closed and cleaned but some requests are pilling up, there is a new error called `ClosedInnerError` to indicate this. It should only happen if someone is doing multi-threading without proper resources management, nonetheless much better than returning nil as it was previously done.

close https://github.com/appsignal/rdkafka-ruby/issues/260

### DIFF
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -27,9 +27,11 @@ module Rdkafka
     def close
       return if closed?
       ObjectSpace.undefine_finalizer(self)
-      @native_kafka.with_inner do |inner|
+
+      @native_kafka.synchronize do |inner|
         Rdkafka::Bindings.rd_kafka_consumer_close(inner)
       end
+
       @native_kafka.close
     end
 
--- a/lib/rdkafka/error.rb
+++ b/lib/rdkafka/error.rb
@@ -92,4 +92,10 @@ module Rdkafka
       super("Illegal call to #{method.to_s} on a closed admin")
     end
   end
+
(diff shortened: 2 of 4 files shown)


mensfeld:
I had zero crashes with this in karafka and waterdrop BUT I am pointing out that I had zero crashes even without any mutexes in rdkafka-ruby as I do have granular locking and optimized shutdown in karafka.

thijsc:
Well we got a green build, so that's good! Retrying it to see if it's green again.

mensfeld:
One thing I wonder is whether we should have one more decrement lock (different to avoid deadlocks) but I was not able to prove that it is needed. Tried my best but decrement would not cause any issues even once

thijsc:
Merging!

mensfeld:
@thijsc you mind crafting a beta? :pray:

thijsc:
Done, `0.13.0.beta.8` is out

PR #277, opened 2023-08-29
Possible race condition in NativeKafka#close

peterzhu2118:
Since the check to closed? is not protected by the mutex, it's possible for two threads to return false for closed? and thus call rd_kafka_destroy twice.

### DIFF
--- a/lib/rdkafka/native_kafka.rb
+++ b/lib/rdkafka/native_kafka.rb
@@ -83,9 +83,9 @@ module Rdkafka
     end
 
     def close(object_id=nil)
-      return if closed?
-
       synchronize do
+        return if closed?
+
         # Indicate to the outside world that we are closing
         @closing = true
 


peterzhu2118:
Hmmm, on second thought, this might not actually fix the issue. There's this code:

```ruby
return unless @inner

Rdkafka::Bindings.rd_kafka_destroy(@inner)
@inner = nil
```

So we should never call `rd_kafka_destroy` on `@inner` twice.

mensfeld:
@peterzhu2118 yes, this should never be called UNLESS you close it from two threads which is not what you should be doing as this part of the API is not thread-safe.

peterzhu2118:
> this part of the API is not thread-safe

I'm not sure what you mean here. The `close` method is wrapped in a `synchronize` block, so it should be thread-safe right?

mensfeld:
@peterzhu2118 you are right and funny thing, as it was actually me who implemented this mutex :facepalm: 

Nonetheless, I do not think this will solve your problem in any way.

PR #282, opened 2023-08-31
Pull in producer partition count cache from karafka-rdkafka

ttstarck:
Related to this PR I opened a while ago https://github.com/appsignal/rdkafka-ruby/pull/230

I am interested in pulling in the fixes that were made in [karafka-rdkafka](https://github.com/appsignal/rdkafka-ruby/commit/ffe84585890d425b6cbdd02020d7f87ca230fa15) regarding the partition caching on the Producer class!

I've cherry-picked those changes over into this PR!

### DIFF
--- a/lib/rdkafka/producer.rb
+++ b/lib/rdkafka/producer.rb
@@ -5,6 +5,11 @@ require "objspace"
 module Rdkafka
   # A producer for Kafka messages. To create a producer set up a {Config} and call {Config#producer producer} on that.
   class Producer
+    # Cache partitions count for 30 seconds
+    PARTITIONS_COUNT_TTL = 30
+
+    private_constant :PARTITIONS_COUNT_TTL
+
     # @private
     # Returns the current delivery callback, by default this is nil.
     #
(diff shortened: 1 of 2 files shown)


mensfeld:
@ttstarck I plan to move it over at some point, hopefully, this year, with a more significant set of changes, and I do have a few remarks about this:

1. Can you use a patch instead of a non-direct cherry-pick? I would love to have my copyrights tracked and for them to remain in between those repos.
2. This patch for Karafka is suited for karafka needs. In the case of "generic" librdkafka I would envision that this is configurable, and there have been some discussions on expanding the configuration beyond only librdkafka. I still did not wrap my head around that, but as I said, I do plan a bigger investment and, hopefully merger of both versions (karafka and this) in the near future.

That said, you will have to give me a bit of time to think about how we could approach it to make the whole ecosystem aligned.

ttstarck:
@mensfeld 
> Can you use a patch instead of a non-direct cherry-pick? I would love to have my copyrights tracked and for them to remain in between those repos.

Fair, I'm not sure how to exactly do that as it seems the [original commit](https://github.com/karafka/karafka-rdkafka/pull/4/commits/fd9f72120abd1a2ef97a46c02b7565b5ece3ecae) is no longer available? When I attempt to apply the merge commit patch it fails...

 >That said, you will have to give me a bit of time to think about how we could approach it to make the whole ecosystem aligned.
 
 Cool, I saw all the changes in the recent release and was worried this may have been missed / forgotten. I'm ok with this waiting a bit longer until y'all have come up with a better solution.
 
 > I do plan a bigger investment and, hopefully merger of both versions (karafka and this) in the near future.
 
 That would be awesome!

mensfeld:
@ttstarck works for me: https://github.com/karafka/karafka-rdkafka/commit/fd9f72120abd1a2ef97a46c02b7565b5ece3ecae.patch

> Cool, I saw all the changes in the recent release and was worried this may have been missed/forgotten.

Let me reassure you that my long-term plan is to merge those two. There are, however some differences between them that make it maybe not hard but complicated from the perspective of their philosophy.

`karafka-rdkafka` can make certain assumptions because it is tied to karafka. Because of that, I can make assumptions about offset storage, timeouts and other things. `rdkafka-ruby` is suppose to be an agnostic pipeline between librdkafka and Ruby realm so I cannot "just assume" what Karafka users want because the scope of operations is broader.

That said I did not forget, I do plan to merge it but I will close this PR and apply it via a patch as said, hopefully this year with bigger piece of work to move this gem forward.

PR #291, opened 2023-09-28
Introduce Ruby 3.2, 3.3.0-preview2 + update older Ruby versions in CI

mensfeld:
This PR updates Ruby versions against which we run the CI.

close https://github.com/appsignal/rdkafka-ruby/issues/290


mensfeld:
Oooo :( semaphore cannot into 3.3.0 :(

mensfeld:
@thijsc can I switch to actions? it would be a bit more convenient to have all in one place for this lib imho and 3.3.0 would work. I have code for this in karafka-rdkafka ready

PR #294, opened 2023-10-12
Replace semaphore with github actions + add Ruby 3.3.0

mensfeld:
close https://github.com/appsignal/rdkafka-ruby/issues/290


mensfeld:
@thijsc as we discussed, can we unplug semaphore once that is green?

thijsc:
I've removed the Semaphore project, looking at the webhooks it looks this should now be done.

PR #296, opened 2023-10-13
Allow stopping a consumer without closing it

webgago:
Let's say we run consumer.each_batch and process messages asynchronously.  At some point the consumer terminates. 
I'd like to stop polling, wait until messages got processed, store offsets and commit. In other words gracefully shutting down. 
Until now, the only option I have to stop iterating is to close consumer. But in this case I can't do final commit.
Please let me know if I'm missing something. Thank you.

### DIFF
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -25,6 +25,7 @@ module Rdkafka
     # Close this consumer
     # @return [nil]
     def close
+      stop
       return if closed?
       ObjectSpace.undefine_finalizer(self)
 
@@ -33,6 +34,14 @@ module Rdkafka
       end
 
       @native_kafka.close
+    end
+
+    def stop
+      @stop_requested = true
+    end
+
+    def stopped?
+      closed? || @stop_requested
     end
 
     # Whether this consumer has closed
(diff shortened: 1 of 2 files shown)


mensfeld:
@thijsc assigning you also here because I have an internal conflict:

on one side this is something that users may benefit from but IMHO, it slightly goes beyond the "API proxy / thin functionality" scope of this library. It does not add any new librdkafka interactions, but it just expands on the simple consumer, thus it could be easily implemented by each of the users using the official apis (polling and looping). Adding more complex state management would require a complex notion of state that is already available in things like karafka or racecar.

webgago:
@mensfeld Thank you for reviewing! 

I just want to add that I'd like to use `each_batch` in my custom consumer implementation. But I can't do it properly and am forced to copy/paste the internals of `each_batch` logic which is something I'd like to avoid.

mensfeld:
I totally understand your point of view. IMHO the consumer iterators themselves should be moved to examples of usage of this library and not to the library itself. State management can be tricky and that is something beyond the scope of this library. Today we add stop, tomorrow we have to add error tracking for exactly-once semantic (which I am working on), etc. And then we end up with APIs we need to officially support.

@thijsc stated several times and I agree with him, that this library aims to be a ruby-friendly complete proxy of librdkafka functionalities. Going with that spirit, we should just have `examples/` illustrating such use-cases but not officially support higher level APIs such as this one. This is exactly what librdkafka does.

mensfeld:
@webgago batch iterator is def. out of scope of rdkafka-ruby as it is a higher level concept.

webgago:
Thanks @mensfeld. That totally makes sense then. @thijsc Feel free to close this PR.

PR #321, opened 2023-10-31
bump librdkafka to 2.2.0

mensfeld:
karafka-rdkafka operates on it without any issues (aside from known).

full changeset:

# librdkafka v2.2.0

librdkafka v2.2.0 is a feature release:

 * Fix a segmentation fault when subscribing to non-existent topics and
   using the consume batch functions (#4273).
 * Store offset commit metadata in `rd_kafka_offsets_store` (@mathispesch, #4084).
 * Fix a bug that happens when skipping tags, causing buffer underflow in
   MetadataResponse (#4278).
 * Fix a bug where topic leader is not refreshed in the same metadata call even if the leader is
   present.
 * [KIP-881](https://cwiki.apache.org/confluence/display/KAFKA/KIP-881%3A+Rack-aware+Partition+Assignment+for+Kafka+Consumers):
   Add support for rack-aware partition assignment for consumers
   (#4184, #4291, #4252).
 * Fix several bugs with sticky assignor in case of partition ownership
   changing between members of the consumer group (#4252).
 * [KIP-368](https://cwiki.apache.org/confluence/display/KAFKA/KIP-368%3A+Allow+SASL+Connections+to+Periodically+Re-Authenticate):
   Allow SASL Connections to Periodically Re-Authenticate
   (#4301, started by @vctoriawu).
 * Avoid treating an OpenSSL error as a permanent error and treat unclean SSL
   closes as normal ones (#4294).
 * Added `fetch.queue.backoff.ms` to the consumer to control how long
   the consumer backs off next fetch attempt. (@bitemyapp, @edenhill, #2879)
 * [KIP-235](https://cwiki.apache.org/confluence/display/KAFKA/KIP-235%3A+Add+DNS+alias+support+for+secured+connection):
   Add DNS alias support for secured connection (#4292).
 * [KIP-339](https://cwiki.apache.org/confluence/display/KAFKA/KIP-339%3A+Create+a+new+IncrementalAlterConfigs+API):
   IncrementalAlterConfigs API (started by @PrasanthV454, #4110).
 * [KIP-554](https://cwiki.apache.org/confluence/display/KAFKA/KIP-554%3A+Add+Broker-side+SCRAM+Config+API): Add Broker-side SCRAM Config API (#4241).

## Enhancements

 * Added `fetch.queue.backoff.ms` to the consumer to control how long
   the consumer backs off next fetch attempt. When the pre-fetch queue
   has exceeded its queuing thresholds: `queued.min.messages` and
   `queued.max.messages.kbytes` it backs off for 1 seconds.
   If those parameters have to be set too high to hold 1 s of data,
   this new parameter allows to back off the fetch earlier, reducing memory
   requirements.

### General fixes

 * Fix a bug that happens when skipping tags, causing buffer underflow in
   MetadataResponse. This is triggered since RPC version 9 (v2.1.0),
   when using Confluent Platform, only when racks are set,
   observers are activated and there is more than one partition.
   Fixed by skipping the correct amount of bytes when tags are received.
 * Avoid treating an OpenSSL error as a permanent error and treat unclean SSL
   closes as normal ones. When SSL connections are closed without `close_notify`,
   in OpenSSL 3.x a new type of error is set and it was interpreted as permanent
   in librdkafka. It can cause a different issue depending on the RPC.
   If received when waiting for OffsetForLeaderEpoch response, it triggers
   an offset reset following the configured policy.
   Solved by treating SSL errors as transport errors and
   by setting an OpenSSL flag that allows to treat unclean SSL closes as normal
   ones. These types of errors can happen it the other side doesn't support `close_notify` or if there's a TCP connection reset.

### Consumer fixes

  * In case of multiple owners of a partition with different generations, the
    sticky assignor would pick the earliest (lowest generation) member as the
    current owner, which would lead to stickiness violations. Fixed by
    choosing the latest (highest generation) member.
  * In case where the same partition is owned by two members with the same
    generation, it indicates an issue. The sticky assignor had some code to
    handle this, but it was non-functional, and did not have parity with the
    Java assignor. Fixed by invalidating any such partition from the current
    assignment completely.

close https://github.com/karafka/rdkafka-ruby/issues/255


bitemyapp:
One clarification I wanted to offer as the person that worked on rebasing and testing @edenhill's `fetch.queue.backoff.ms` patch is that it isn't exactly about memory requirements. The issue is that without that setting, you can't use increased buffer capacity to increase total consumer throughput beyond a certain point. The default backoff for the fetch queue means increasing buffer just increases the time between fetches in a way that keeps the throughput functionally constant.

What I was trying to do and thankfully @edenhill's patch enabled me to achieve was:

- Fix the Consumer group sub-linear scaling (throughput scaled linearly in my testing after 3 consumers)
- Make the fetch rate tuneable so that I could drive Consumer throughput harder at the expense of some additional CPU utilization.
- Make the Consumers (fetch queue) drive lagged partitions harder so that there was less partition lag skew due to uneven distribution of sleepy and busy partitions. The more the consumption rate is limited by I/O and CPU and the less it's limited by the fetch queue backoff timeout, the less of a problem partition lag skew becomes.

Partition lag skew is a problem that becomes worse as you add more instances to the Consumer group because you're averaging out the variation across fewer partitions. It kills your efficiency when trying to scale up the consumer group by adding more instances/consumers.

I'm processing ~7+ GiB/second of ~1-1.5 KiB messages that require parsing, analysis, and metrical tracking. I wanted to be able to do so without having to deploy 100+ instances. Currently it's ticking along with 32 instances utilizing ~5 CPU cores (10 worker threads) each with head-room to spare. The data is Zstd compressed.

mensfeld:
@bitemyapp I copy-pasted original release notes from librdkafka. Do you want me to include something else?

bitemyapp:
No that's fine, I was just expounding on it. Didn't mean to distract you here, just wanted to leave some summary notes.

mensfeld:
You did not distract me at all :) thank you for this note. I plan to do extensive work in this lib soon (actually already started) so tips like this will help me improve the docs as well :pray:

PR #328, opened 2023-11-02
Update librdkafka location reference

mensfeld:
This PR changes repo org reference from edenhill to confluentinc because of the location transition.


nijikon:
👍

PR #353, opened 2023-11-22
Add admin#delete_group utility

piotaixr:


mensfeld:
@piotaixr one of the specs seem to be off. Can you correct it? :pray: otherwise LGTM and I think it could go with the upcoming 0.15.0 release that I plan for the next week.

piotaixr:
This is really weird. The first time the test runs on an empty kafka cluster (`docker compose down -v && docker compose up -d`), it fails. The second time it works.

I don't know how to fix that in a non-ugly way...

mensfeld:
Ping if you need help :)

mensfeld:
@piotaixr test order will not be deterministic because rspec runs them in random. I will look into this so don't worry :)

piotaixr:
> @piotaixr test order will not be deterministic because rspec runs them in random. I will look into this so don't worry :)

Are you sure RSpec orders by random as default? I see that the first spec group to be executed on my laptop is always
```
Rdkafka::AbstractHandle
  A subclass that does not implement the required methods
    raises an exception if operation_name is called
    raises an exception if create_result is called
  .register and .remove
    should register and remove a delivery handle
  #pending?
    when true
      should be true
    when not true
      should be false
  #wait
    should wait until the timeout and then raise an error
    when not pending anymore and no error
      should return a result
      should wait without a timeout
    when not pending anymore and there was an error
      should raise an rdkafka error
```

mensfeld:
@piotaixr will check that tops tomorrow :)

mensfeld (approved):
Going to merge it.

Lets deal with the theoretical order issue if it occurs. I was not able to reproduce it either while running on a random seed.

mensfeld:
@piotaixr since you are in this code area, wouldn't you be interested in implementing also the list group capabilities? :pray:

PR #363, opened 2023-12-08
Nix support

alexandriainfantino:


mensfeld:
@alexandriainfantino thanks for this PR. I will take a look at it in the upcoming week :pray:

ext/Rakefile
```diff
@@ -1,40 +1,48 @@
-# frozen_string_literal: true
```
mensfeld, line 1: was there any problem with the literal here?

mensfeld (commented):
@alexandriainfantino would you mind providing a bit of explanation here? I do understand the context but it would be good to have a justification in the code itself

ext/Rakefile
alexandriainfantino: nope, can add back in!

alexandriainfantino:
@mensfeld Basically nix can't locate the file paths because the packages it's requiring aren't managed by the system but are managed by nix itself, so using the normal file paths doesn't work for nix users.

Mini_portile causes an issue because it's dependencies are downloaded on the fly and therefore don't exist/aren't accessible in the nix environment

mensfeld:
@alexandriainfantino would you mind adding this explanation to the code in the relevant section? That way we accumulate knowledge in the codebase :) I have minimal exp with Nix, thus I will rely on you in that matter.

ext/Rakefile
```diff
@@ -1,40 +1,55 @@
...
 
 require File.expand_path('../../lib/rdkafka/version', __FILE__)
-require "mini_portile2"
 require "fileutils"
 require "open-uri"
 
 task :default => :clean do
-  # Download and compile librdkafka
-  recipe = MiniPortile.new("librdkafka", Rdkafka::LIBRDKAFKA_VERSION)
+  if ENV.fetch('RDKAFKA_EXT_PATH', '').empty?
```
geoff2k, line 8: Usage-wise, it looks like will be up to the invoker to set `RDKAFKA_EXT_PATH`, is that correct?  So instead of running what the doc suggests:

```
cd ext && bundle exec rake && cd ..
```

They should run something like this under Nix:

```
cd ext && RDKAFKA_EXT_PATH=/some/path/user/provides/ bundle exec rake && cd ..
```

Is that correct?

mensfeld: @alexandriainfantino would you mind ^^

alexandriainfantino: Hey!

`RDKAFKA_EXT_PATH` is a part of our nix configuration, it will be up to the invoker to ensure their nix configuration is set up appropriately

geoff2k: Could we add a few lines to the documentation indicating what needs to be done in a Nix configuration?  Or maybe just add a comment here so that if someone searches the codebase for "Nix" they might get a hit?

I'm not trying to make a mountain out of a molehill, but it feels like if we are going to add support for a particular environment, we should make it as easy as possible for users of that environment to know we support it and how best to use it.

Not trying to lead the witness, but maybe something like:

```
raise "RDKAFKA_EXT_PATH must be set in your environment when running under nix" unless ENV["IN_NIX_SHELL"]
```

alexandriainfantino: Absolutely! will add

alexandriainfantino: pushed up a small check with comment 👍

thijsc (approved):

nijikon (approved):

ext/Rakefile
mensfeld: @geoff2k you good? Wanted to include this in the upcoming release and also backport to karafka-rdkafka (I do my best to sync them and retire karafka-rdkafka)

mensfeld:
@alexandriainfantino going to give @geoff2k few more days and will merge and will release alongside of other things in the upcoming release. I will also backport this to karafka-rdkafka.

mensfeld:
Ok. I don't see any other comments so will merge.

mensfeld (approved):

mensfeld:
@alexandriainfantino thank you! I will schedule this for the upcoming release and will backport to karafka-rdkafka until both are merged.

PR #368, opened 2023-12-11
Replace rd_kafka_offset_store with rd_kafka_offsets_store

mensfeld:
librdkafka docs on rd_kafka_offset_store:

> This API lacks support for partition leader epochs, which makes it at risk for unclean leader election log truncation issues. 

this PR replaces deprecated and unsafe `rd_kafka_offset_store` with stable and recommended `rd_kafka_offsets_store`

The internal structure is slightly different but he API remains unchanged.

It is worth pointing out, that this API CANNOT be used when `enable.auto.offset.store` and will end up with an error, but this method always had this indicator:

```ruby
# When using this `enable.auto.offset.store` should be set to `false` in the config.
```

so I will not provide a backward compatibility fallback, as it should never be used when a consumer is automatically committing offsets.

close https://github.com/karafka/rdkafka-ruby/issues/256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Rdkafka Changelog
 
 ## 0.15.1 (Unreleased)
+- [Enhancement] Replace `rd_kafka_offset_store` with `rd_kafka_offsets_store` (mensfeld)
 - [Enhancement] Alias `topic_name` as `topic` in the delivery report (mensfeld)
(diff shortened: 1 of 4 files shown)


mensfeld:
@thijsc added you to the PR to have a third pair of :eye: as this is IMHO a big change. No API changes but I want to be sure it won't cause any issues as I don't know why you picked `rd_kafka_offset_store` instead of `rd_kafka_offsets_store` in the first place.

nijikon (approved):
LGTM 🚀

PR #382, opened 2023-12-27
improve report building

mensfeld:
This is a backport from karafka-rdkafka that handles a case where a failed report (one with error) would contain error that did not have a topic because the message did not reach it. I do not recall an exact case where it was happening but I do know it did.


nijikon (commented):
LGTM

PR #409, opened 2024-02-05
Lazy initialize logger thread for forks

mensfeld:
it does NOT change fundamentally how logger works but fixes it and provides fork support.

additionally due to lazy-loading as long as there are no logs (which is not that rare), this thread will not be initialized.

close https://github.com/karafka/rdkafka-ruby/issues/407
backport: https://github.com/karafka/karafka-rdkafka/pull/72

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Rdkafka Changelog
 
+## 0.15.2 (Unreleased)
+- [Fix] Background logger stops working after forking causing memory leaks (mensfeld)
+
 ## 0.15.1 (2024-01-30)
 - [Enhancement] Provide support for Nix OS (alexandriainfantino)
 - [Enhancement] Replace `rd_kafka_offset_store` with `rd_kafka_offsets_store` (mensfeld)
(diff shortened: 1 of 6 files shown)


mensfeld:
@ioquatix @KJTsanaktsidis :pray: also please if you guys don't mind

KJTsanaktsidis (commented):
Looks good to me 👍 Libraries can also use `Kernel._fork` patching for this kind of thing but it's much of a muchness.

I do wonder though, what's the point of bouncing the logging through a queue at all? Are logger implementations not required to be threadsafe?

mensfeld:
> I do wonder though, what's the point of bouncing the logging through a queue at all? Are logger implementations not required to be threadsafe?

They are but this will be a subject to a separate PR to redesign this.

thijsc:
> I do wonder though, what's the point of bouncing the logging through a queue at all? Are logger implementations not required to be threadsafe?

This was implemented because there were some crashes in the past related to logging when exiting a process. Maybe this is more reliable with newer versions of librdkafka.

mensfeld:
For now I will leave it as it is as I plan to move a lot of this code to events-based flow and then I will revisit this thread existence.

@thijsc I will keep an eye on anything of that nature. Thanks!

PR #417, opened 2024-03-01
switch to use mutex in AbstractHandle.wait method

nijikon:
Closes https://github.com/karafka/rdkafka-ruby/issues/390

### DIFF
--- a/lib/rdkafka/abstract_handle.rb
+++ b/lib/rdkafka/abstract_handle.rb
@@ -15,6 +15,9 @@ module Rdkafka
     # Registry for registering all the handles.
     REGISTRY = {}
 
+    # Default wait timeout is 31 years
+    WAIT_TIMEOUT = 10_000_000_000
+
     class << self
       # Adds handle to the register
       #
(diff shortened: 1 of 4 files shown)


lib/rdkafka/abstract_handle.rb
```diff
@@ -45,7 +54,7 @@ def pending?
     # on the operation. In this case it is possible to call wait again.
     #
     # @param max_wait_timeout [Numeric, nil] Amount of time to wait before timing out.
-    #   If this is nil it does not time out.
+    #   If this is nil we will use WAIT_TIMEOUT value
```
mensfeld, line 57: "If this is nil, will wait forever" would be more explicit IMHO. WAIT_TIMEOUT is an implementation detail

lib/rdkafka/callbacks.rb
```diff
@@ -157,6 +157,8 @@ def self.process_create_topic(event_ptr)
           create_topic_handle[:error_string] = create_topic_results[0].error_string
           create_topic_handle[:result_name] = create_topic_results[0].result_name
           create_topic_handle[:pending] = false
+
+          create_topic_handle.broadcast
```
mensfeld, line 161: can we name it "unlock" or something similar? broadcast seems outside of our domain

mensfeld (changes requested):

mensfeld (approved):

nijikon:
@thijsc will you have time this coming week to look at this change?

thijsc (approved):
This looks good to me.

PR #421, opened 2024-03-17
fix unlock race condition

mensfeld:
This PR ensures that we change the handler status inside a mutex when we unlock. Otherwise there could be an intermediate state where a critical section would not catch the unlock


nijikon (commented):
Great find! 🎉

PR #424, opened 2024-03-22
Allow for disabling regex ext during build

mensfeld:
Currently librdkafka by default uses libc regex engine. It is somehow weird not supporting things like `\d\d` and other matches. While it seems you can always bypass that users are confused. I had several questions about it already.

This PR adds a flag that allows for disabling of the external engine in favour of local one that seems to match the Ruby one.

It os **not** turned on by default to have backwards compatibility.

### DIFF
--- a/ext/Rakefile
+++ b/ext/Rakefile
@@ -27,6 +27,14 @@ task :default => :clean do
       :sha256 => Rdkafka::LIBRDKAFKA_SOURCE_SHA256
     }
     recipe.configure_options = ["--host=#{recipe.host}"]
+
+    # Disable using libc regex engine in favor of the embedded one
+    # The default regex engine of librdkafka does not always work exactly as most of the users
+    # would expect, hence this flag allows for changing it to the other one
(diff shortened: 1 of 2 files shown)


nijikon (approved):

ext/Rakefile
```diff
@@ -27,6 +27,14 @@ task :default => :clean do
       :sha256 => Rdkafka::LIBRDKAFKA_SOURCE_SHA256
     }
     recipe.configure_options = ["--host=#{recipe.host}"]
+
+    # Disable using libc regex engine in favor of the embedded one
+    # The default regex engine of librdkafka does not always work exactly as most of the users
+    # would expect, hence this flag allows for changing it to the other one
+    if ENV['RDKAFKA_DISABLE_REGEX_EXT']
```
nijikon, line 34: Maybe `ENV.key?('RDKAFKA_DISABLE_REGEX_EXT')`?

nijikon (approved):
LGTM

PR #426, opened 2024-03-22
Introduce instance initialization callback

mensfeld:
This PR introduces "yet another callback" so as not to break the current APIs.

Context: in case a producer or consumer, after creation, they immediately poll. This is expected, especially since the producer and admin use a background thread started automatically but this creates a chicken-and-egg issue where, for a multi-instance setup, there is no way to initialize and register the appropriate oauth handlers before first polling but after creation of rdkafka instance.

This is needed because otherwise, we get a callback with a name for which we did not register a handler. In theory, this could be handled by dynamically registering a new handler and injecting it, reserving the instance, but this is not good for any complex system where callbacks may be more advanced.

This callback aligns with the current clumsy way of doing things that should be replaced (and will be) by a proper instrumentation flow to ensure we do not break any APIs. It also matches other callback behaviors and the order of callable arguments.

Not perfect but gets things done within the current approach.


nijikon (approved):
LGTM

mensfeld:
superseeded by https://github.com/karafka/rdkafka-ruby/pull/427

PR #428, opened 2024-03-23
Allow for deferred native kafka start

mensfeld:
This PR tackles https://github.com/karafka/rdkafka-ruby/pull/426 in a different way.

Instead of yielding, we allow for deferred start of producer/consumer/admin.

That way we can obtain full rdkafka-ruby object on which we can operate without resolving to the bindings and we can do complex initialization and only then actually start the requested operations.

This is much better than the yielding attempt for higher level frameworks that should not resolve to using bindings API directly but should rely on rdkafka-ruby abstractions.

This change is NOT breaking.

Note: I do NOT provide "not_started_check" in a similar manner to how we do close producer check, etc because this is a low-level API that is suppose to be used only by higher level libs devs.


nijikon (commented):
LGTM

PR #437, opened 2024-04-18
Fix early return in NativeKafka#start

ferrous26:
I think this was a typo, and could potentially lead to creating multiple polling threads.

I'm investigating a segfault that I see in production during shutdown where `rd_kafka_poll` is crashing because `@inner` appears to be `nil`.


mensfeld:
Great catch! Though you should not run from master :man_singer: 

:pray:

PR #439, opened 2024-04-23
Remove .0 versions of Ruby for CI

mensfeld:
I do not see any reason to maintain it that way. While I can understand lack of ability to upgrade to higher easily, I do not understand this within same version.


nijikon (approved):
LGTM

PR #440, opened 2024-04-25
stabilize spec

mensfeld:
This sleep is needed as the describe changes do not catch fast enough after creating ACLs and sometimes actions fail and our integration server for 24/7 spec execution.

### DIFF
--- a/spec/rdkafka/admin_spec.rb
+++ b/spec/rdkafka/admin_spec.rb
@@ -288,6 +288,9 @@ expect(ex.broker_message).to match(/Topic name.*is invalid: .* contains one or m
         expect(create_acl_report.rdkafka_response).to eq(0)
         expect(create_acl_report.rdkafka_response_string).to eq("")
 
+        # Since we create and immediately check, this is slow on loaded CIs, hence we wait
+        sleep(2)
+
         #describe_acl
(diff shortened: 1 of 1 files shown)


spec/rdkafka/admin_spec.rb
```diff
@@ -288,6 +288,9 @@
         expect(create_acl_report.rdkafka_response).to eq(0)
         expect(create_acl_report.rdkafka_response_string).to eq("")
 
+        # Since we create and immediately check, this is slow on loaded CIs, hence we wait
+        sleep(2)
```
nijikon, line 292: Can we move this to a method that will check if we are on a CI and sleep only then?

mensfeld: No because it can also happen outside of CI once in a while. I would prefer stability over conditions.

nijikon: Sounds good. Can you adjust the comment cause it is misleading. Maybe move this to a method so we won't have to add sleeps every where and just call that method?

mensfeld: ref https://github.com/karafka/karafka/issues/2041

PR #441, opened 2024-04-25
Fix bogus case/when syntax. Levels 1, 2, and 6 previously defaulted to UNKNOWN

jjowdy:
Ruby will evaluate expressions like `0 || 1 || 2` to the first truthy value before matching the `case`, so only level 0 would ever match the first case, etc.


mensfeld:
Great catch!

mensfeld (approved):

PR #444, opened 2024-04-29
Promote to 0.16.0

mensfeld:
I'm moving to `0.16.0` since @nijikon removed `wait_timeout`, and we added a warning. It should not go as a patch release.


nijikon (approved):
LGTM

PR #445, opened 2024-04-30
Provide errors describing Admin API

mensfeld:
This PR introduces error describing API that we can use to get librdkafka errors detailed data. It can be useful for debugging.

This API does not need connection to Kafka nor an admin instance, that's why it is a class level one.


nijikon (approved):
LGTM

PR #449, opened 2024-05-06
Producer support for per-topic configs

mensfeld:
This PR introduces the ability to use topic-specific custom settings as defined here: https://karafka.io/docs/Librdkafka-Configuration/ on a per-message dispatch basis.

This allows one producer to have different `acks` levels specific to topics and other custom settings.

In addition, it caches the rdkafka topic object so we do not have to create it with each request, saving on memory allocations. Since `with_inner` is locking, we do not have to worry about race conditions when adding configs to the cache.

We use the config `hash`  value, so the same config created from different Ruby hash objects will reuse the same cache entry.

We allow multiple configs per the same topic because I've stumbled upon a case where the user had different importance messages delivered to the same topic and wanted to tune this. Makes sense to me.

close https://github.com/karafka/rdkafka-ruby/issues/436


spec/rdkafka/producer_spec.rb
```diff
@@ -31,6 +31,46 @@
...
+      it 'expect to raise error' do
+        expect do
+          producer.produce(topic: 'test', payload: '', topic_config: { 'acks': 1 }).wait
+        end.not_to raise_error
+      end
+
+      context 'when alteration should change behavior' do
+        # This is set incorrectly for a reason
+        # If alteration would not work, this will hang the spec suite
+        let(:producer) { rdkafka_producer_config(
```
nijikon, line 53:
```ruby
let(:producer) do
  rdkafka_producer_config(
    'message.timeout.ms': 1_000_000,
    :"bootstrap.servers" => "localhost:9094",
  ).producer
end
```

nijikon (approved):
LGTM

lib/rdkafka/producer.rb
```diff
@@ -54,6 +59,48 @@ def initialize(native_kafka, partitioner_name)
...
+
+        return if @topics_configs[topic].key?(config_hash)
+
+        # If config is empty, we create an empty reference that will be used with defaults
+        rd_topic_config = if config.empty?
+                            nil
+                          else
+                            Rdkafka::Bindings.rd_kafka_topic_conf_new.tap do |topic_config|
+                              config.each do |key, value|
+                                error_buffer = FFI::MemoryPointer.from_string(" " * 256)
```
KJTsanaktsidis, line 83: Can you do `FFI::MemoryPointer.new(:char, 256)` to avoid actually having to call `String#times` here?

lib/rdkafka/producer.rb
```diff
@@ -54,6 +59,48 @@ def initialize(native_kafka, partitioner_name)
...
+
+                                unless result == :config_ok
+                                  raise Config::ConfigError.new(error_buffer.read_string)
+                                end
+                              end
+                            end
+                          end
+
+        @topics_configs[topic][config_hash] = config
+        @topics_refs_map[topic][config_hash] = Bindings.rd_kafka_topic_new(inner, topic, rd_topic_config)
```
KJTsanaktsidis, line 100: I verified that this does not leak `rd_topic_config` even if `rd_kafka_topic_new` fails - `rd_kafka_topic_new` destroys the config even if the function errors (e.g. https://github.com/confluentinc/librdkafka/blob/a4bf4876fadff054cf729fcb26d0d31e93cff44b/src/rdkafka_topic.c#L299-L304)

KJTsanaktsidis: That said: does this need to raise an exception here in the case where `rd_kafka_topic_new` fails? Otherwise it'll just write `nil` to `@topics_refs_map[topic][config_hash]`

lib/rdkafka/producer.rb
mensfeld: Technically `read_string` is not the same here but you are probably right that it does not matter. Fixing

thijsc (approved):
Looking good!

lib/rdkafka/producer.rb
mensfeld: Yeah good point. It should not write the nil. I took it "errorless" because it used to be like this. Let me alter that.

mensfeld: Ok we won't fix it here becasue `rd_kafka_errno2err` is not exposed. But I can at this stage (will create a separate issue) raise generic exception when this returns null to prevent from it leaking out.

mensfeld: ref https://github.com/karafka/rdkafka-ruby/issues/451

mensfeld: Raising this here. In other places a subject to refactor.

PR #456, opened 2024-06-10
improve object allocations

mensfeld:
This PR reuses the existing topic name in the handler to build reports. This allows us not to read from FFI and not create new objects.

We also use `bytesize` instead of `bytes.size` that saves one array allocation.


nijikon (approved):
LGTM

PR #462, opened 2024-06-19
Add `seek_by` method

zinahia:
This is to enable using only the needed parameters and remove the need to provide or create a fake `Rdkafka::Consumer::Message` which is an implementation wrapper that should not be needed for the API of the clients when calling `seek`.

It is my humble opinion that providing this alternative can give a cleaner interface for the clients to use, however I'm happy to receive feedback or change it in case there is something I might be missing. 

The specs I added are based on the existing specs of the `seek` method.

PS to the maintainers: Thanks for your work, this is a great library ❤️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.17.0 (Unreleased)
 - [Enhancement] Update `librdkafka` to `2.4.0`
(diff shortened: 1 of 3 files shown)


mensfeld (changes requested):
You mind adding this to the changelog?

mensfeld:
Thanks. This change aligns well with the nature of the lib (low level) as it actually provides lower level API than the one already existing :)

mensfeld (approved):

PR #526, opened 2024-11-22
Avoid clobbering LDFLAGS and CPPFLAGS if in a nix prepared environment

secobarbital:
Support for nix was added in #363, but the gem still does not compile for users with both nix AND homebrew. In that situation, the Rakefile is still prioritizing homebrew. A better test for nix is the presence of the NIX_LDFLAGS env var, which is present if and only if nix has prepared the environment to use libraries and headers in the nix store.


ext/Rakefile
```diff
@@ -16,9 +16,9 @@ task :default => :clean do
     require "mini_portile2"
     recipe = MiniPortile.new("librdkafka", Rdkafka::LIBRDKAFKA_VERSION)
 
-    # Use default homebrew openssl if we're on mac and the directory exists
-    # and each of flags is not empty
-    if recipe.host&.include?("darwin") && system("which brew &> /dev/null") && Dir.exist?("#{homebrew_prefix = %x(brew --prefix openssl).strip}")
+    # Use default homebrew openssl if we're on mac and the directory exists, is not using nix-prepared libraries
+    # and each of flags is not already set
+    if recipe.host&.include?("darwin") && system("which brew &> /dev/null") && Dir.exist?("#{homebrew_prefix = %x(brew --prefix openssl).strip}") && !ENV["NIX_LDFLAGS"]
```
mensfeld, line 21: can you use `key?` instead of the `!ENV` syntax pretty plz? :pray:

mensfeld (changes requested):

ext/Rakefile
secobarbital: I assume you want the other two checks to be changed to.

mensfeld: @secobarbital if you want but it is not required. It's ok to change it here only as well if it's too much hustle for you. Thank you

mensfeld (approved):

PR #564, opened 2025-03-26
allow for customized native kafka poll timeout

mensfeld:
close https://github.com/karafka/rdkafka-ruby/issues/552

### DIFF
--- a/lib/rdkafka/config.rb
+++ b/lib/rdkafka/config.rb
@@ -233,11 +233,12 @@ module Rdkafka
     #
     # @param native_kafka_auto_start [Boolean] should the native kafka operations be started
     #   automatically. Defaults to true. Set to false only when doing complex initialization.
+    # @param native_kafka_poll_timeout_ms [Integer] ms poll time of the native Kafka
     # @return [Producer] The created producer
     #
     # @raise [ConfigError] When the configuration contains invalid options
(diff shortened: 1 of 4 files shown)


lib/rdkafka/config.rb
```diff
@@ -233,11 +233,12 @@ def consumer(native_kafka_auto_start: true)
     #
     # @param native_kafka_auto_start [Boolean] should the native kafka operations be started
     #   automatically. Defaults to true. Set to false only when doing complex initialization.
+    # @param native_kafka_poll_timeout_ms [Integer] ms poll time of the native Kafka
     # @return [Producer] The created producer
     #
     # @raise [ConfigError] When the configuration contains invalid options
     # @raise [ClientCreationError] When the native client cannot be created
-    def producer(native_kafka_auto_start: true)
+    def producer(native_kafka_auto_start: true, native_kafka_poll_timeout_ms: 100)
```
geoff2k, line 241: 👋 

Curious where the default value of 100 came from?  Also since we re-use this default, would it make sense to pull it into a constant?  Happy to submit a PR for either/both.

mensfeld: @geoff2k the 100 came from the previous default. I did not change it and it was there from beginning.

Having it in a constant makes sense. Happy to accept the PR :pray:

PR #626, opened 2025-06-24
Revert "remove each_batch method due to data consistency issues (#541)"

natetarrh:
This reverts commit dc3bd4fafdcc2571f1be48d8b80b0dae0c8b89a0.

`#each_batch` is a useful primitive. For Kafka consumers in general, duplicate processing during consumer group rebalances is inevitable in certain situations and exactly once consumption is not guaranteed. I believe a recommendation that a consumer implementation be idempotent is more appropriate than directing users to Karakfa.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Rdkafka Changelog
 
 ## 0.21.1 (Unreleased)
+- **[Breaking]** Restore `#each_batch`.
 - [Enhancement] Allow for producing to non-existing topics with `key` and `partition_key` present.
 - [Enhancement] Replace TTL-based partition count cache with a global cache that reuses `librdkafka` statistics data when possible.
 - [Enhancement] Support producing and consuming of headers with mulitple values (KIP-82).
--- a/lib/rdkafka/consumer.rb
+++ b/lib/rdkafka/consumer.rb
@@ -609,23 +609,87 @@ module Rdkafka
       end
     end
 
-    # Deprecated. Please read the error message for more details.
+    # Poll for new messages and yield them in batches that may contain
+    # messages from more than one partition.
+    #
+    # Rather than yield each message immediately as soon as it is received,
+    # each_batch will attempt to wait for as long as `timeout_ms` in order
+    # to create a batch of up to but no more than `max_items` in size.
+    #
+    # Said differently, if more than `max_items` are available within
+    # `timeout_ms`, then `each_batch` will yield early with `max_items` in the
+    # array, but if `timeout_ms` passes by with fewer messages arriving, it
+    # will yield an array of fewer messages, quite possibly zero.
+    #
+    # In order to prevent wrongly auto committing many messages at once across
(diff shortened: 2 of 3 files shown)


mensfeld:
The removal of `#each_batch` isn't only about duplicates - that's expected Kafka behavior. The core issue is **ordering guarantees**.

Kafka guarantees messages are processed in order within each partition. However, `#each_batch` violated this by allowing batches to complete processing out of order after rebalances:

1. Consumer starts receiving data
1. Rebalance occurs
1. Consumer receives rewinded data
1, Pre-rewind (newer) data is present **before** older data

This breaks Kafka's fundamental ordering contract, which consumers rely on. The primitive was incompatible with Kafka's ordering model because it allowed batches to "leak" across rebalance boundaries without proper cleanup.

While `#each_batch` seemed useful, it created a footgun that could silently violate ordering in ways that would be hard to debug. Proper batch handling requires explicit rebalance cleanup to maintain Kafka's core ordering guarantees.

Also:

> This library does not intend to offer complex producers or consumers. The aim is to stick closely to the functionalities provided by librdkafka itself.

While I agree `#each_batch` conceptually is useful, its implementation here is not complete and fundamentally flawed.

natetarrh:
Thanks for the detailed reply and clarification that it's an ordering concern.

I'm not sure there is an ordering problem, though. How is this different from a consumer that fails to commit an offset (for example, because it's shutting down) which then starts reprocessing from the previously committed offset? It's not an ordering violation because the consumer is reprocessing messages that it has already seen. For a batch A, B, C, D where the consumer has processed A and B but is rewound to offset A, it may process C and D before it processes A, B, C, and D again, but since it has already processed A and B, continuing to process C and D do not violate the ordering guarantee (or the consumer no longer holds the partition it held previously, so it doesn't matter).

I understand and agree with the desire to avoid any complex consumer implementations. I'm just hoping to make a case for `each_batch` not being fundamentally flawed as written, since implementers of Kafka consumers should expect to handle cases like this (reprocessing).

mensfeld:
Hey Nate,

> I'm not sure there is an ordering problem, though

There is. I spent considerable time debugging this issue with a user under extremely heavy load on a low-latency networking setup. It **is** 100% reproducible and **does** happen in production.

The issue is when and how batches get yielded to user code with partially accumulated data crossing the rebalance barrier. The violation isn't about the consumer's internal state - it's about what gets presented to user code. 

Kafka's ordering guarantee means "if you consume partition messages, you see them in offset order." But `each_batch` can yield batches where messages are out of offset order due to rebalance timing.

You're absolutely right that robust consumers should handle reprocessing, but they shouldn't have to reason about whether messages within a single batch might be from different sides of a rebalance boundary. That's the kind of subtle complexity that leads to bugs.

The primitive would need rebalance callbacks and complex state management to be safe, which moves it beyond the "wrapper around librdkafka API" scope I aim for in this karafka ecosystem component. Since this is a real need, I've implemented proper batch handling with full rebalance safety in Karafka's OSS version.

PR #874, opened 2026-05-07
Optimize Headers.from_native: pre-allocate scratch pointers, dedup names

mensfeld:
- Use String#-@ to deduplicate and freeze header name strings, reducing GC pressure for repeated header keys across messages.


mensfeld:
closing. no real performance benefit here and we cannot pre allocate because of jruby s upport.

PR #964, opened 2026-09-02
Close native clients before Ruby shutdown

selesse:
I ran into a shutdown SIGSEGV where ffi can `dlclose` librdkafka during CRuby's forced finalization while its native threads are still running. The next thread to execute jumps into unmapped librdkafka code. I first chased this in https://github.com/ffi/ffi/pull/1194, but ffi can't know which client owns those threads or call `rd_kafka_destroy` for it.

There are two gaps on this side. `rd_kafka_new` starts those threads before the Ruby wrapper and its finalizer exist, so a later setup failure can orphan the native handle. Successfully-created clients also currently rely on GC finalizers during interpreter teardown.

The first commit makes construction exception-safe and destroys any handle that later setup fails to wrap. The second weakly tracks completed clients so `at_exit` closes them while Ruby can still stop and join their threads. Weak references preserve normal GC behavior, and the existing PID guard keeps inherited handles with their parent after a fork.


mensfeld:
@selesse can you demonstrate this behaviour with an integration stress test?

mensfeld:
I have somewhat similar code stashed locally so please give me some time to figure out if yours can superseed mine :pray:  (likely few days)

mensfeld:
Ok mine is for a different case so I think I can mary them both. Integration demo is appreciated though :pray:

lib/rdkafka/config.rb
```diff
@@ -343,6 +335,36 @@ class NoLoggerError < RuntimeError; end
...
+          opaque: opaque,
+          auto_start: false,
+          timeout_ms: timeout_ms
+        )
+        client = yield(kafka, native_kafka)
+        native_kafka.start if auto_start
+        Rdkafka::ClientCleanup.register(client)
+      rescue Exception
+        owner = client || native_kafka
+        owner ? owner.close : Rdkafka::Bindings.rd_kafka_destroy(kafka)
```
mensfeld: if this fails we will never reach `raise` hence shadowing the original error

i think we should at least log such cases

lib/rdkafka/client_cleanup.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# Main Rdkafka namespace of this gem
+module Rdkafka
+  # Tracks live public clients so they can be closed before Ruby begins forced finalization.
+  # @private
+  module ClientCleanup
```
mensfeld: this actually tracks clients (registry) and manages shutdown so maybe we could just call it `Clients`?

mensfeld (changes requested):

selesse:
> can you demonstrate this behaviour with an integration stress test?

I added an integration test as a separate commit. :)

I made it deterministic instead of using a conventional stress test. The original shutdown race only crashes about one run in ten, which felt too flaky for CI, so this test forces the unsafe shutdown order in a child process instead.

On master it crashes 10/10. With this PR, the control still crashes 10/10 while normally-created clients survive 10/10. That confirms the test can reproduce the bug and that the cleanup prevents it.

It runs on Linux/glibc and skips platforms where this shutdown behaviour doesn't occur. Let me know if you prefer I write it another way!

mensfeld:
@selesse amazing. Thank you. I should be able to take care of this today and craft a release tomorrow.

mensfeld:
FYI I will NOT bump librdkafka with this so users who pin librdkafka (for whatever reasons) can benefit out of this. After that I will try to craft an update with librdkafka as well.

mensfeld:
We still support Ruby until the end of this month so I added a proper Ruby 3.2 fix.

mensfeld (approved):

mensfeld:
Thank you. I will do my best to tackle one more PR and to craft a release.

